PR #8, opened 2017-04-21
Refactor App container

amberleyromo:
- Pulled out some of the main tab functionality in App/index.js into separate components to start to make that file more readable. Added under /components because they are more presentational, most logic getting handed down.


shayc:
this is great

src/components/Keyboard.js
```diff
@@ -0,0 +1,31 @@
...
+  render() {
+    return (
+      <div className="keyboard">
+        <textarea ref={ref => { this.textarea = ref }} placeholder="Type some text"></textarea>
+        <RaisedButton
+          label={this.props.intl.formatMessage({ id: 'cboard.containers.Text.speak' })}
+          labelPosition="before"
+          icon={volumeUpIcon}
+          primary={true}
+          onTouchTap={(event) => { event.preventDefault(); this.props.speak(this.textarea.value) }}
```
agupta93, line 19: dont create arrow functions here a new function will be created at every render so if pure render is used in Raised buttton then it will be of no use.

src/components/NavigationBar.js
```diff
@@ -0,0 +1,50 @@
...
+const boardIcon = <FontIcon className="material-icons">view_module</FontIcon>;
+const keyboardIcon = <FontIcon className="material-icons">keyboard</FontIcon>;
+
+class NavigationBar extends PureComponent {
+  render() {
+    return (
+      <Paper zDepth={1}>
+        <BottomNavigation selectedIndex={this.props.selectedIndex}>
+          <BottomNavigationItem
+            label={this.props.intl.formatMessage({ id: 'cboard.containers.App.bottomNavigationItem.settings' })}
```
agupta93, line 18: can we cache this.props to a variable? as `this` is not minified while creating bundle

PR #13, opened 2017-05-14
fix: add const to terms definition to prevent hoisting terms variable…

mystetskyivlad:
fix: add const for terms variable to prevent hoisting to global


shayc:
Thanks Vlad :)

PR #39, opened 2017-08-22
Speak symbols 'text' property if provided instead of 'label'.

### LINKED ISSUE #33: symbol should speak label or text
Currently when you click on a symbol it speaks the content of the label property of symbol.
We should make changes so when a text property has a value, that value will get precedence over the label value when using speech synthesis.

The text value was planned specifically to override the label when using speech.

Should look something like this:

```
speakSymbol() {
  const text = symbol.text || symbol.label;
  this.speak(text);
}
```

falinsky:
Fixes #33


shayc:
Thanks!

PR #42, opened 2017-08-24
Prevent body scroll when dragging symbols

### LINKED ISSUE #32: Prevent body scroll when dragging symbols
on iOS devices, when rearranging symbols the body also scrolls, preventing fluid dragging of symbols.
can try overflow: hidden on body when dragging.. not sure

falinsky:
Fixes #32


shayc:
I'm going to try the other workaround, adding another event handler just for iOS should be the last resort.
Will report back

shayc:
Apparently this bug does not reproduce on iOS 11, closing.
Thanks for the PR though!

Issue #37, opened 2017-08-21
General <Snackbar /> Component


shayc:
Need to create a general (app wide) Snackbar component, to display various app-wide messages.
Should use [material-ui snackbar](https://material-ui-1dab0.firebaseapp.com/demos/snackbars/)

hoodwink73:
Hey, @shayc I am giving this a go. I will try to be thoughtful with the interface. And post it here, shortly. Let me know if you have any specific requirements.

shayc:
Sounds awesome, the Snackbar component should be general, it should use Redux for state and have a 'message': String and 'open': Boolean properties to show/hide the Snackbar.. If you need help with Redux or have any other questions I am happy to help.

You'll need to create:

- Redux reducer
- Redux action creators
- Redux actions
- Container Snackbar that uses [material-ui](https://material-ui-1dab0.firebaseapp.com/demos/snackbars/) snackbar
- Snackbar.propTypes = { open: PropTypes.boolean, message: PropTypes.string }

hoodwink73:
I am on the right path.

I am also thinking of building a simple queue for the notifications.

`<SnackBar />` suggests showing only one notification at a time. I am respecting this guideline. 

So, suppose notification B triggers when notification A is already displayed.
I would wait until notification A gets dismissed or autohidden, after which I will display notification B.

Do you know think this notification queue will be helpful here?

shayc:
Good point, I checked [material design guidelines](https://material.io/guidelines/components/snackbars-toasts.html#snackbars-toasts-usage) - just to see what the spec says:

> Consecutive snackbars
> 
> There is only ever one snackbar displayed. When a second snackbar is triggered while the first is displayed, the first should start the contraction motion downwards before the second one animates upwards.

Your suggestion is on track with the spec, notification queue will be helpful in some scenarios but if it takes too long to implement we can do with a basic Snackbar for now

PR #46, opened 2017-08-25
Introduce a simple notifications system

hoodwink73:
This system lays the groundwork to show basic notifications by dispatching actions. 

Only one notification is shown at a time as per the [Material UI Guidelines](https://material.io/guidelines/components/snackbars-toasts.html#snackbars-toasts-usage)

The system still does not handle multiple notifications triggered at once. Right now, it will simply replace the message in place.

Addresses #37


hoodwink73:
Hey @shayc, just wanted to get your idea on the progress.

This does not include queuing multiple notifications. We can discuss some ideas for that. Post that we can use this to actually notify some events and write some test cases.

Here is a [quick tour](https://www.dropbox.com/s/dutwvgm8rzyy1sm/Kapture%202017-08-25%20at%2011.00.04.mp4?dl=0) of me dispatching the notifications using Redux Dev Tools

src/containers/Notifications/Notifications.js
```diff
@@ -0,0 +1,67 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import Snackbar from 'material-ui/Snackbar';
+import { hideNotification as hideNotificationActionCreator } from './actions';
+
+class Notifications extends Component {
+  constructor(props) {
+    super(props);
+    this.handleNotificationDismissal = this.handleNotificationDismissal.bind(
```
shayc, line 10: can avoid .bind and use property assignment that will have 'this' in scope

handleNotificationDismissal = ()=>{}

shayc (approved):
good job, thanks!

src/containers/Notifications/Notifications.js
hoodwink73: Hmmm. Sure will do.

I got informed [from this](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md). Don't know whats the right thing to do now. But I think having a consistent style is more important.

shayc: The link is referring to having .bind / arrow functions inside JSX props or anywhere in the render() method, which will indeed create a new function everytime render() is called.
What you did (and also what I offered) are essentially the same in that they will not create a new function for each render() call since they are defined at the class level. It’s just more readable in my opinion.

PR #49, opened 2017-08-25
Edit multiple symbols

### LINKED ISSUE #25: Edit multiple symbols
We want to enable the user to edit existing symbols.
Right now the app enables the user to select multiple symbols but doesn't allow editing.

**How the feature should work:** once selecting the symbols (using the SELECT button), the user will be able to click on an "edit" button which will open a SymbolsDetails componet dialog, this component will present one or more symbols for edit, if there are multiple symbols, the component should be paginated using [material-ui stepper](https://material-ui-1dab0.firebaseapp.com/component-demos/stepper), only one symbol per page, the user will be able to navigate between the symbols by clicking on "next" or "previous" buttons. When the user submits the dialog all changes will be saved (redux), if the user cancel, changes will be canceled.

ioRekz:
fixes #25

### DIFF
--- a/src/containers/Board/Board.js
+++ b/src/containers/Board/Board.js
@@ -6,6 +6,7 @@ import classNames from 'classnames';
 import Button from 'material-ui/Button';
 import IconButton from 'material-ui/IconButton';
 import DeleteIcon from 'material-ui-icons/Delete';
+import EditIcon from "material-ui-icons/Edit";
 import CheckCircleIcon from 'material-ui-icons/CheckCircle';
 import AddBoxIcon from 'material-ui-icons/AddBox';
 import SettingsIcon from 'material-ui-icons/Settings';
@@ -16,7 +17,8 @@ import {
   previousBoard,
   addBoard,
   addSymbol,
-  deleteSymbols
+  deleteSymbols,
+  editSymbols
 } from './actions';
 import speech from '../../speech';
 import messages from './messages';
@@ -152,7 +154,7 @@ export class Board extends Component {
   };
 
   handleAddClick = () => {
-    this.setState({ symbolDetailsOpen: true });
+    this.setState({ symbolDetailsOpen: true, selectedSymbols: [], isSelecting: false })
   };
 
   handleEditClick = () => {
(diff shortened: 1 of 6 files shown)


src/containers/Board/Board.js
```diff
@@ -152,7 +154,7 @@ export class Board extends Component {
   };
 
   handleAddClick = () => {
-    this.setState({ symbolDetailsOpen: true });
+    this.setState({ symbolDetailsOpen: true, selectedSymbols: [], isSelecting: false })
```
ioRekz, line 157: I think it's better to cancel the selecting mode when adding a new symbol.
Several times I tried to play the newly added symbol without noticing I was still in select mode

src/containers/Board/SymbolDetails/SymbolDetails.js
```diff
@@ -138,28 +186,19 @@ export class SymbolDetails extends Component {
...
-    type: 'symbol'
-  }
+  editingSymbols: []
 };
 
 const mapStateToProps = state => {
-  return {
-    selectedSymbols: state.board.selectedSymbols
-  };
+  return {};
```
ioRekz, line 197: We don't need anything from the store. selectedSymbols wasn't use anyway.
I don't know if it can stays like that or we should just not connect this component

src/containers/Board/SymbolDetails/messages.js
```diff
@@ -5,4 +5,8 @@ export default defineMessages({
     id: 'cboard.containers.Board.SymbolDetails.addSymbol',
     defaultMessage: 'Add Symbol',
   },
+  editSymbol: {
+    id: 'cboard.containers.Board.SymbolDetails.editSymbol',
+    defaultMessage: 'Edit Symbol'
```
ioRekz, line 10: This could be a plural "edit symbols" but since this message already existed from previous version I would stick with it. We are editing symbol one by one anyway

src/containers/Board/SymbolDetails/SymbolDetails.js
```diff
@@ -15,55 +16,78 @@ export class SymbolDetails extends Component {
...
+  }
+
+  updateEditingSymbol = (id, property, value) => state => {
+    const editingSymbols = state.editingSymbols.map( s =>
+      s.id === id ? {...s, ...{ [property]: value} } : s
+    )
+    return {...state, editingSymbols}
+  }
+
+  updateNewSymbol = (property, value) => state => {
```
ioRekz, line 46: used the functional setState so it's possible to update 2 fields without one overwriting the other

src/containers/Board/Board.js
shayc: totally agree!

shayc:
THIS. IS. AWESOME!

shayc (commented):
looking good, thanks!

src/containers/Board/SymbolDetails/SymbolDetails.js
shayc: I'll check later if we can remove connect

src/containers/Board/SymbolDetails/messages.js
shayc: for now we can leave it as is.. this can be addressed later down the road when we find some UX designer

PR #51, opened 2017-08-25
fix: add id to created symbols

### LINKED ISSUE #47: After adding new symbols, selecting one of them select all the others
It happens because added symbols don't have ids so they all resolve to undefined and get selected.

ioRekz:
fixes #47

### DIFF
--- a/src/containers/Board/reducer.js
+++ b/src/containers/Board/reducer.js
@@ -20,7 +20,7 @@ function symbolReducer(board, action) {
   switch (action.type) {
     case ADD_SYMBOL:
       return Object.assign({}, board, {
-        symbols: [...board.symbols, action.symbol]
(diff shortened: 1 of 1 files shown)


src/containers/Board/reducer.js
```diff
@@ -20,7 +20,7 @@ function symbolReducer(board, action) {
   switch (action.type) {
     case ADD_SYMBOL:
       return Object.assign({}, board, {
-        symbols: [...board.symbols, action.symbol]
+        symbols: [...board.symbols, Object.assign({}, action.symbol, { id: board.symbols.length })]
```
shayc, line 23: Hey, good catch, I've been meaning to add unique ids, I will implement with this package: [shortid](https://www.npmjs.com/package/shortid) soon

shayc (approved):
thanks you!

PR #52, opened 2017-08-25
undefined export boards issue 

jquintozamora:
add boards redux integration within backup to solve issue export undefined


src/containers/Settings/Settings.js
```diff
@@ -94,7 +104,12 @@ export class Settings extends Component {
           onSubmit={this.goBack}
         />
         <Speech open={this.state.speechOpen} onCancel={this.goBack} />
-        <Backup open={this.state.backupOpen} onCancel={this.goBack} />
+        <Backup
+          boards={this.props.boards}
```
shayc, line 108: we usually destruct props in the beginning of the render() function

shayc (approved):
good fix thanks 👍

PR #53, opened 2017-08-25
Debounce sample message on settings change

BrendanFDMoore:
This PR debounces the speech sample message
so it doesn't queue a stack of calls on rapid settings changes.

Fix for #40


shayc (approved):
good refactor of speakSample and debounce

shayc:
Thanks for this and sorry about the duplicate debounce, I wrote you in the Discord channel but should've DM you instead!

BrendanFDMoore:
That's ok - sorry was out of town for the weekend so couldn't tidy up the conflicts. Thanks for handling that!

Yes, I think I prefer this version which reduces the repeated destructuring of `intl` from `this.props` and isolates it close to final usage.

👍

PR #55, opened 2017-08-27
Any notifications dispatched during an already active notification is queued

hoodwink73:
Adds more capability to #37


hoodwink73:
I give a small [demo here](https://www.dropbox.com/s/v4pic0gv94gbolf/Kapture%202017-08-27%20at%2021.25.07.mp4?dl=0). To be more explicit I have added a voice narration.

shayc:
that was fast :P good job!

Issue #48, opened 2017-08-25
Keyboard accessibility


AlmeroSteyn:
Became aware of this awesome project from a tweet by Dan Abramov.

Had a look at it today and really awesome effort!!

I found a couple of (what I think could be) easily fixable issues that would improve the keyboard accessibility of the app.

First and foremost is the use of the following style in `Board.css`:

```
.Symbol:focus {
  outline: 0;
}
```

This, of course, removes the focus outline from tiles, making keyboard navigation pretty much impossible. 

Is this something you want to address in the app? If so I want to see if I can get around to setting up a PR to address basic keyboard accessibility.

shayc:
Hi and welcome :) making the app accessible is a top priority, you're invited to make PR with a fix, let me know if you need help!

AlmeroSteyn:
Thank you! A little busy this side on day job things, will get to it as soon as I can.

a11yup-old:
Also became aware through Dan on Twitter. Really cool project. Keep up the good work 👍 

I'd also love to help out with improving keyboard accessibility for the app.
@AlmeroSteyn Let me know if I can help out here ;) I have a little bit of time a the moment.

AlmeroSteyn:
@dakeyras7 Help would be great! Sorry it took so long, wanted to take the time to have an initial opinion, which only happened today. Please comment on the pull request with your ideas.

Opened PR #61

shayc:
I've fixed the styling with the IconButton on focus, I'll have to refactor later to abstract IconButton implementation, probably into another Component or via theme configuration

shayc:
The more I think of the problem the more I'm inclined to have main keyboard navigation be done by arrow keys and backspace/delete keys. Tab navigation should be enabled to traverse all interactive elements but not be used as the main nav method since it is cumbersome. Please be aware that once changing the symbols layout within the app, tab navigation gets messed up (the DOM/taborder doesn't change). So this needs further implementation. For now we have the basics - thank you!

AlmeroSteyn:
@shayc Thanks for merging the PR and picking up the Iconbutton issue!

Maybe some points for consideration on the navigiation method. Users who make use of keyboard navigation, especially those using assistive tech, often have a specific expectation of how this works.

[...]

PR #61, opened 2017-09-01
Add keyboard accessibility

AlmeroSteyn:
Here is a pull request for issue #48. 

Here are some initial adjustments to add basic keyboard accessibility to the app.

1. Activate the summary bar with enter as well and improve focus highlighting.
2. Changed the background colour of the header bar so that the focus outlines of the buttons become visible.
3. Add focus outline to the button tiles.

The second change obviously changes the style of the bar but as material ui is used, which I am unfamiliar with, this was the quickest way to demonstrate what I have in mind. Perhaps it is possible to achieve a similar inverted result with the black bar?

It would also, I think, be great to add keyboard mechanisms (skip links, maybe keyboard shotcuts) to quickly reach the various sections. 

Focus control still needs to be fixed. For example, when you navigate to another section of speech tiles, focus is lost to the top of the page again, which is nasty for keyboard users.

Perhaps we could discuss these here first to see which way you would like to go? 

@dakeyras7 would love help on this! Just needed some time to have a good look at it and have something sensible to say about it first :-)

PS: I am seeing some deprecation warnings on the console with regards to React.createClass and using PropTypes from inside React itself.  Have not tracked it down where it is coming from but seems to be inside some dependency. This will block the upgrade to React 16.


src/containers/Board/Board.css
```diff
@@ -42,7 +42,9 @@
 }
 
 .Symbol:focus {
-  outline: 0;
+  border-width: 5px;
+  border-color: blue;
+  border-style: solid;
```
a11yup-old, line 47: How about keeping the `outline: 0` in, because you are kind of replacing it with the border anyway, right? Else we get both the border + the browser-dependent outline.

Also, what do you think about a box-shadow to highlight a focused element with a sort of hover effect? This also would not make the tile contents jump.
Since the app is quite colorful already, I would also suggest a neutral grey color for this.

So, in effect something like this:

```css
.Symbol:focus {
  outline: none;
  box-shadow: 0px 0px 20px 5px #333;
}
```

a11yup-old:
Regarding the keyboard navigation in general. I still see two main issues, which you also mentioned briefly:

1. focus on the tiles-area is lost once we are selecting a "folder"
2. navigation shortcuts would be nice. without them the user has to tab through a lot of items first 
 
Regarding the first one. I could take a stab at it @AlmeroSteyn, if you don't object ;)
Not sure how to solve the second one yet. I'm thinking "quick-row-selection" or similar (with numbers maybe?).

src/containers/Board/Board.css
shayc: totally agree with the comment

shayc:
Good job so far, I wouldn't change the ToolBar's background color, instead I assume we can change the individual IconButton on focus/hover styles, we're using [material-ui](https://material-ui-1dab0.firebaseapp.com/customization/overrides/#overriding-with-classes) and it's using CSS in JS. I can happily do the color changes.

Also we can do a hidden a11y menu, that will appear on the first tab click, that will have a 2 shortcuts, one for Output element, another for the Board symbols.

shayc:
If you can revert the color changes to the Toolbar I can merge, and we can continue working, I can also do it if you wish as there are some conflicts with current version

shayc (approved):
Thanks! merging with some fixes I'll add soon, just don't want it to be harder to merge while we develop new features

src/containers/Board/Board.css
AlmeroSteyn: Hi. Sorry I was AFK this weekend :-)

Agreed I should have added the outline 0.

Some nitpicks:

The reason I added such a glaring border was that I see the app as an accessibility tool, therefore I would normally accentuate a11y features in such an app. In this case for users with low vision who are not yet using assistive tech to browse.

The box shadow version is a little lower in contrast, but in my opnion not so that it will prove a problem, although I would probably have cranked up the contrast just a little more.

What I do think should be considered is that `border` has better browser support than `box-shadow`. For example, IE 9 upwards supports `box-shadow`, which means that an IE 8 user will experience a totally inaccessible keyboard experience again. Now this is mitigated by the fact that React also only support IE9 upward, but regardless, using `border` insures a larger group of browsers will display this correctly.

AlmeroSteyn:
Thanks for the merge @shayc! I was AFK this weekend.

src/containers/Board/Board.css
a11yup-old: @AlmeroSteyn 
Yeah, the contrast could be even higher. Agreed :)

You're right regarding the support for `box-shadow`: We could use graceful degradation here. I think there is some "old trick" for IE<9 and the rest would just pick up `box-shadow`. Would have to look it up again though.

AlmeroSteyn: @dakeyras7 Hey as long as the solution is as widely supported as the browser outline itself, be-it from a fallback or not, it's got my vote :)

PR #64, opened 2017-09-03
Feat/issue 57 symbol details

amberleyromo:
wip - put up early at @shayc request


shayc (approved):
thanks, tidy! :) gonna move the render method to the end of the component to align with the rest of the code, let me know if you wanna change convention, also will move the symbols property out of the state

amberleyromo:
I would be interested to discuss convention-- I like it toward the top because i almost feel like it serves as a pseudo-table of contents. You can get a glance without having to scroll all the way to the bottom. Thoughts?

shayc:
I like the ui implementation to be at the bottom, but I agree with you regarding birds eye view when entering a component, how about we put the propTypes and defaultProps as static members or const at the very top. Can serve as a mini documentation

amberleyromo:
Consistency is most important, that's fine. Just wondering why you like it at the bottom?

shayc:
Well, the community tends to organize React component methods in the order they're being called:
Airbnb React/JSX Style Guide: [Ordering](https://github.com/airbnb/javascript/tree/master/react#ordering)

The styleguide also lists static methods first, so we can define `propTypes`, `defaultProps` first thing, what do you say?

PR #65, opened 2017-09-04
Symbol Focus Tracking

a11yup-old:
This PR addresses further improvements to the keyboard navigation.

With this, users are able to navigate "down" a folder and when they return with the back link the focus will be restored back to that folder.
For a quicker (more convenient?) navigation backwards I added a listener for the `Esc` key. It feels like a good choice to me usability-wise. I am not 100% sure though if this is a bad idea for screen-readers. @AlmeroSteyn: Do you have an opinion on this?

I decided not to do "Skip-Links" for now and focused on the focus tracking itself first.

Implementation details: Every time a Symbol is focused, I keep track of it being the lastly focused Symbol on that specific board. Also I am using React `ref`s to store references to the Symbol DOM elements, so that I can focus them programmatically via `focus()`. The latter happens on `componentDidUpdate` of the Board component.


AlmeroSteyn:
Great start and ideas!

With using `Esc` to nav back up the board chain: I don't see a problem with this kind of shortcut. It is used already for closing modals, etc, so this does not strike me as crazy and won't mess with screen readers. 

One may want to check later if this context change is recognised well enough by screen readers, but this can be solved with ARIA Live Regions. If it comes to this I wrote a plugin recently to easily provide ARIA live messaging in a React app if this is something that could be helpful: https://www.npmjs.com/package/react-aria-live. I found that ARIA live regions behave erratically in React apps if used far down in the component tree and this package solves that by rendering it super high up in the tree.  But perhaps it is best to defer these typed of additions to a separate PR in the future?

So long some general comments. They don't refer to specific lines perse so I will summarise it here:
1. Perhaps consider creating a ref to the `Symbol` component itself and then focussing the internal element in a focus function of the component? Dunno if passing an internal element through to the parent of the component separates the child component well enough. While focussing the `Symbol` component itself will leave the actual focus implementation to the component and its internal implementation itself. For this to happen the `Symbol` component will need to become a class component as it has no instance at the moment.

So in `Symbol` one could set the `ref` internally to the button
```
 <button
      className={symbolClasses}
      ref={button => this.symbolButton= button}
      onClick={() => {
        onClick(symbol);
      }}
      onFocus={() => onFocus(symbol)}
    >
```

 and then create a `focus` handler on the converted component class: 
```
focus() {
    this.symbolButton.focus();
}
```

Then in `Board.js` one can set a ref to the component itself:

```
 <Symbol
            {...symbol}
            ref={element => this.symbolElements[key] = element}
            onClick={this.handleSymbolClick}
            onFocus={this.handleSymbolFocus}
          >
```

Which can then be focussed in the code by:
`this.symbolElements[key].focus();`

Then the parent can worry about focussing `Symbol` components and the `Symbol` component can worry about how that focus is applied to itself.

What do you think?

2. Initially I was worried about the specific tiles losing focus, as they are now doing in my running example. If I set focus to the 10th tile on the board and press enter, focus is reset to the first tile on the board. I did a quick debug and indeed the first element is still stored as last focussed and refocussed every time. We will need to ensure that the currently selected tile retains focus even if the tiles are re-ordered for some reason.

Your idea to focus the folder tile when navigating back is a great idea, but I am almost tempted to think that if we go to this level it should be stored in Redux with the current board state. This state seems to be stored in localstorage and then rehydrated into the Redux store on refresh. So adding it to the Redux implementation would then conserve this functionality on refresh, otherwise it will only be available for levels I have visited in THIS session and will break halfway through drilling back up as the internal component state has been lost.

Your ideas on this? 

Would also love to hear @shayc take on this as this would impact the Redux architecture.

3. Check missing semi colons om some lines :-)

@shayc Is `Prettier` already part of the project? I saw some mention of this made somewhere.

shayc:
Quick reply here: Prettier is set up on my local machine, but no [configuration](https://github.com/prettier/prettier#configuration-file) has been set, I'll integrate soon so we can have consistent PRs.
Will write a full reply soon, thanks!

a11yup-old:
Thanks for your feedback. Much appreciated 👍 

Let me answer in order:

1.  You are right. My implementation breaks the separation between child and parent. Great suggestion you made there, and I would like to incorporate it. But as you said that this turns the `Symbol` component from a functional to a class component. I don't mind that, but would be curious what @shayc thinks about it.

2a. I am not completely sure what you mean. Before this PR selecting a tile would reset the focus to the first tile on the board, correct. With this PR the focus stays on the selected tile on the board. Or did I miss something? :)

2b. I agree with you. Ideally I would have moved this into the redux store, but I didn't want to mess with the store architecture in this first step and thought this would suffice. We could do it though. Also curious what @shayc thinks about this.

3. Will run my code through prettier locally and make those fixes. Thanks for spotting that ;)

shayc:
@AlmeroSteyn Thank you for the detailed replies! I would love the focus state to be saved in the Redux store.

@dakeyras7 Regarding 1. I am starting to think we should divide `Symbol` into two components, the first one a simple template `Symbol` functional component with some styles, the second `BoardButton` as class component that will handle the `onClick` and `onFocus` events and also contain its relevant styles. The reason is re-usability, we have UI similar to `Symbol` in the `Output` component, right now it's duplicate code. If you don't mind I'll take the seperation task and let you know so you can implement above.

shayc:
Pushed a refactor, we now have two components `Symbol` `BoardButton` as described above. Please pull version, thanks! I'm also thinking we'll need to save scroll position to the Redux store, perhaps a different PR?

a11yup-old:
Thanks @shayc. Let me pull your refactor and adapt my changes.

Regarding the storage of the focus in the redux store: Let me see if I can come up with a solution within this PR.

a11yup-old:
I put the currently focused button by board into the redux store. The value from there is now used to determine if a button should get its focus restored. I went a more declarative way this time. The `BoardButton` has a new prop `hasFocus` that is determined by  the `Board` component during rendering. Then inside `BoardButton` the component itself checks this prop on `didMount` and `didUpdate` to claim its focus.

The reason I went this way is because it eliminates the need to store any `ref`s inside `Board` which seemed a bit cleaner to me. The only component that stores a DOM ref now is `BoardButton` which is fine, because it needs to call `.focus()` on it.

src/containers/Board/Board.js
```diff
@@ -187,14 +193,28 @@ export class Board extends Component {
     }));
   };
 
+  handleBoardKeyUp = (event) => {
+    if (event.keyCode === 27) {
```
shayc, line 197: I would refactor the key code literal into an enum reference like:
```
const keyCodes = {
  ESCAPE: 27
}
```
There's probably an NPM package for this, either way should be a key-codes.js file that will be imported

shayc (approved):
Great PR, thank you for this!

src/containers/Board/Board.js
shayc: added [keycode] npm package and refactored above :)

Issue #70, opened 2017-09-05
Simplify process for adding/updating translations


BrendanFDMoore:
I'm working on setting up some scripts to help smooth for process of adding and modifying translation strings to Crowd In for approved contributors.

Right now the process is pretty clunky, and I think it can be a lot cleaner. I'm opening an issue just to  share that this is WIP to avoid duplicated effort.

shayc:
[crowdin api](https://support.crowdin.com/api/api-integration-setup/)

BrendanFDMoore:
First PR towards this goal is open #75

PR #75, opened 2017-09-12
Add script to pull latest translations from CrowdIn

BrendanFDMoore:
This PR adds a tool/script to automatically pull the latest translation string for *all* languages directly from CrowdIn via their API. It only updates the local files, so you will need to commit the changes to the repo as usual after running `yarn translations:pull`

The API key for CrowdIn should not be shared publicly, so this PR _also_ adds secrets management to allow us to have encrypted keys in the repo. Very few team members will need access to these keys, and they are not required to run or develop the app.

Future work: Build out the scrip to *push* new translations to CrowdIn.


BrendanFDMoore:
This relates to issue #70

shayc (approved):
clean code, thanks!

Issue #87, opened 2017-10-10
Refactor: Split component & container


BrendanFDMoore:
This is an omnibus issue to track conversion for major components to split the stateless UI `component` from the redux-connect `container`.

To Do:

- [x] Board
- [x] Settings
- [x] Settings: Language
- [ ] Settings: Speech
- [x] Settings: Backup
- [ ] Notification
- [ ] Header
- [ ] Grid
- [ ] ...more?

BrendanFDMoore:
@shayc is doing `Board` and I'm working on `Settings` for now.

BrendanFDMoore:
Now working on the sub-modules of `Settings`: Language, Speech, Backup

shayc:
Added basic container/component separation for `Board` module, marking as done for this dev phase

PR #88, opened 2017-10-10
[Issue 87] Refactor `Settings` container/component

BrendanFDMoore:
This PR aims to refactor only the `Settings` component (not even its children just yet), to split the UI `component` (a pure function of props) portion from the stateful `container` where to connections to the redux store etc reside.

I believe I had encapsulated all of the logic into the container, but please feel free to comment if anything looks amiss.

Relates to #87


shayc (approved):
looking good!

PR #92, opened 2017-10-14
corrected Dutch translations

winnie334:
Fixed all the machine generated errors, which there were surprisingly not that many of. I'm pretty sure I got them all, but if you find anything that looks suspicious, please let me know! I'll do my best to explain why it looks that way or find the correct word if it's incorrect.

Also: I noticed that in the graphical interface the images for left and right click have been switched, perhaps this is something that should be looked at. (symbol.electricalComputer.leftClick & symbol.electricalComputer.rightClick)


shayc:
looks awesome, thanks!
I'll ask the guy responsible on the icons from Straight Street regarding right/left mouse clicks - good catch!

Issue #89, opened 2017-10-11
[Translation] Proof read machine translations


shayc:
We've machine translated 3400 symbols from the Mulberry symbol set into 33 languages.
Most of the languages need proofreading, you can either proofread using Crowdin WYSIWYG interface (recommended) or by manually editing the `.json` files (no visual context - not ideal).

To proofread:

- [Crowdin WYSIWYG interface](https://cboard-org.github.io/cboard-translate/)
- [json files: `src/translations/{language}.json`](https://github.com/cboard-org/cboard/tree/master/src/translations)

Languages proofread:

- [ ] Arabic
- [ ] Bengali
- [ ] Chinese Simplified
- [ ] Czech
- [ ] Danish
- [x] Dutch
- [ ] English
- [x] English, United Kingdom
- [ ] Finnish
- [x] French
- [x] German
- [ ] Greek
- [x] Hebrew
- [ ] Hindi
- [ ] Hungarian
- [ ] Indonesian
- [ ] Italian
- [ ] Japanese
- [ ] Khmer
- [ ] Korean
- [ ] Nepali
- [ ] Norwegian
- [ ] Polish
- [x] Portuguese
- [ ] Romanian
- [x] Russian
- [ ] Sinhala
- [ ] Slovak
- [x] Spanish
- [ ] Swedish
- [ ] Thai
- [ ] Turkish
- [x] Ukrainian
- [ ] Vietnamese

shayc:
@ioRekz took French

Luanmega:
I'll make the Spanish translation

karate:
I'll do the Greek translation (el)

shayc:
This is amazing @karate @Luanmega thanks for making Cboard available in your area!

winnie334:
I'll do the Dutch translation! (nl.json)

Edit: done, see pull request #92. :)

shayc:
@winnie334 thanks a lot, I've approved the translation, pushing to prod :) cheers!

winnie334:
@shayc No problem, glad I could help :)

heyphilllie:
@shayc I just noticed that german has not been proofread yet, so yeah, I could do that or rather would start right away!  :)

shayc:
@l3ekka, thank you!

heyphilllie:
No problem. :-) I just got through page one and two though, there have been some more changes to be made. I could go on tomorrow. If someone else is up for this, just let me know.

btk:
[...]

PR #93, opened 2017-10-16
updated de.json (page 1 and 2)

heyphilllie:
I added some german translations for #89 . So far I've been through page one and two. 

I noticed that the translations in the json itself do not differ between men and women when it comes to job titles. In german the job title depends on the persons gender. Maybe this is something you could think about in the future - or just let me know if I should open an issue for that.


shayc:
You are right, and this is something we need to consider in the near future, it is also a problem in several other languages (not only with job titles). You are more than welcome to open an issue, we can attend this after October ends and we start our second development cycle.

Again thank you so much!

PR #101, opened 2017-10-26
update translation de.json

heyphilllie:
updated german translations in de.json for #89


shayc:
Thank you so much for the hard work!

PR #127, opened 2017-11-20
#goodnessSquad Implemented undo for the board with redux-undo

ktamir:


shayc:
Thanks, just noting, this is still a W.I.P.

PR #130, opened 2017-12-11
fix(translation): Fallback to en-US if the user language has not be translated

arthurdenner:
Hi, @shayc.

When I first tried to run the project, I saw the error below because my language (pt-BR) has not be translated yet (I'll do it soon in Crowdin).

To fix it, I added a catch to the translation import, so when the language isn't found, it fallbacks to en-US and the user is notified.

Let me know if you want to do this different, if you want to change the notification or anything.

**Print of the error:**
image

### DIFF
--- a/src/components/LanguageProvider/LanguageProvider.container.js
+++ b/src/components/LanguageProvider/LanguageProvider.container.js
@@ -74,7 +74,7 @@ export class LanguageProvider extends Component {
         this.setState({ messages });
       })
       .catch(() => {
-        changeLang('en-US');
+        changeLang(DEFAULT_LANG);
         showNotification(`A ${lang} translation was not found, so the language was set to English (en-US).
           Go to Settings if you want to change it.`);
       });
--- a/src/components/LanguageProvider/LanguageProvider.container.js
+++ b/src/components/LanguageProvider/LanguageProvider.container.js
@@ -5,6 +5,7 @@ import { IntlProvider } from 'react-intl';
 
 import { APP_LANGS, DEFAULT_LANG } from '../App/App.constants';
 import { changeLang, setLangs } from './LanguageProvider.actions';
+import { showNotification } from '../Notifications/Notifications.actions';
 import { importTranslation } from '../../i18n';
 
 export class LanguageProvider extends Component {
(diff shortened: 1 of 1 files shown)


src/components/LanguageProvider/LanguageProvider.container.js
```diff
@@ -64,11 +65,19 @@ export class LanguageProvider extends Component {
...
 
-    importTranslation(lang).then(messages => {
-      this.setState({ messages });
-    });
+    importTranslation(lang)
+      .then(messages => {
+        this.setState({ messages });
+      })
+      .catch(() => {
+        changeLang('en-US');
```
shayc: literal string, should use `DEFAULT_LANG` variable instead (equals to 'en-US')

shayc (approved):
Great PR!

src/components/LanguageProvider/LanguageProvider.container.js
arthurdenner: Oops, my bad, didn't see it. I changed it.

shayc:
This is an important catch! I've been meaning to default to the closest related language when no specific language translation is found.

So for example if `pt-BR` isn't available it'll search `APP_LANGS` for a string starting with `pt`, and find the closest available language, in your case `pt-PT` which do exist. I don't know how different (if at all) `pt-PT` is from `pt-BR`. I know for `en-US` and `en-GB` it can be a problem.

Still something to think about.

arthurdenner:
That's an interesting point. We don't know if the user know English at all, so maybe it's a better solution to set a similar language, but we don't know how similar the languages are too.

For example, `pt-PT` and `pt-BR` are like `en-US` and `en-GB`, there are a lot of similarities, but we can't say the same for arabic languages.

shayc:
I'll think about it a little more and let ya know, right now I'm inclined for the similar language solution.
Also I've set up support for `pt-BR` in crowdin.com, also machine translated all strings, so if you want to proofread you're more than welcome.

Theres https://shayc.github.io/cboard-translate or you can just user their website https://crowdin.com/project/cboard

Will merge `pt-BR` support in a bit to `master`

arthurdenner:
Thanks! I'll proofread as soon as I can.

PR #133, opened 2017-12-15
Prevent board button select/deselect on drag and drop

### LINKED ISSUE #109: Board button select/deselect on drag and drop
When dragging and dropping a board button, the button gets selected/deselected.
Correct behavior: button should not be selected/deselected after dropping.

Relevant Componets:
- `BoardButton`
- `Grid`

falinsky:
Fixes #109


shayc (approved):
good PR, I would find a way to not use onDrag handler if possible, from perf perspective. Can do later..

PR #137, opened 2018-01-12
Change to windows EOL in .editorconfig, add recommendations

SteveALee:
I just noticed all files use CRLF line endings so changed editorconfig to match and avoid loads of changes to due to EOLs changing. This area is generally a mess with editor settings and git settings interacting. It might mean non windows devs will have problems. I usually set everything to LF these days.

I also just discovered the nifty VS code extensions recommendations option, so added the key ones to streamline the DX for code users.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,7 +4,7 @@ root = true
 [*]
 indent_style = space
 indent_size = 2
-end_of_line = crlf
+end_of_line = lf
 charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,7 +4,7 @@ root = true
 [*]
 indent_style = space
 indent_size = 2
-end_of_line = lf
+end_of_line = crlf
 charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,13 @@
+# editorconfig.org
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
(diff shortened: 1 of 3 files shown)


SteveALee:
ooh er - it's listing previous commits above - still everything "should" be OK

shayc:
I'll need to check CRLF vs LF on multi-platform projects, I don't want non-windows users to have a problem. I think GitHub automatically converts on push to LF, but still need to check for best practice.

Nice finding on VS Code extensions prop! Will merge soon!

SteveALee:
I personally use linefeeds on windows and configure git appropriately.as I use code with editor config I know no sourcefiles will have CRLF.

All your files have CRLF so unless you accept this there Is a danger someone will save a file changing all the line endings. 

I think you might be able to set the git project CRLF handling rather than relying on user unknown settings. I think it changed since I last investigated the mess.

SteveALee:
So it looks like [little has changed](https://help.github.com/articles/dealing-with-line-endings/) and .gitattributes is still the per repo way to set it up. If you trust git to not knacker binary files the use auto - else use binary.

I'd say that is a separate issue to this PR - I just exposed it for you :)

my preference is `* -text` so git never does any messing with your files and rely on editors to be configured correctly. If you review every PR you'll see any bulk changes due to change in EOL

shayc (approved):
I've removed editor config extension as I think prettier extension does the same, and I would like to install the bare minimum

shayc:
regarding EOL, I also removed the changes, I'm on Windows and have not had any problems of yet (didn't hear anything from Mac/Linux users as well) so I'm going with "if it's ain't broken..."

SteveALee:
There are edge cases depending on editor used and default git for windows config. I was trying to cover as many as possibkle but no doubt the community is not big enough to worry about it.

PR #162, opened 2018-02-20
snapshot tests utils & snapshot test for About component

johanzilber:


shayc (approved):
Looking good! thanks!

Issue #157, opened 2018-02-20
Dockerize the app


shayc:
Create a [Docker ](https://www.docker.com/)container for the app

locnguyen:
If no one has taken this I can do it. I think my only question is why cboard needs a container if it's not a server app.

shayc:
@martinbedouret and @zachi are currently developing the server side, you can see in the app's high-level [roadmap](https://github.com/cboard-org/cboard/blob/master/roadmap.md), and the [cboard-api](https://github.com/cboard-org/cboard-api) repo.

amberleyromo:
@shayc closed with #166?

PR #166, opened 2018-02-28
Dockerize cboard

locnguyen:
I'm using intermediate stages to reduce the image's file size down. Only the runtime dependencies and build artifacts from ﻿`npm run build` are included in the final release stage.

The README is updated with instructions on how to build and run the Docker image.

This is for Issue #157


shayc:
Hurray! 😄 thank you so much!

Issue #171, opened 2018-03-09
Change `npm` calls in CONTRIBUTING.md to `yarn`


chiptus:
First to say that this looks like an amazing project. I've immediately sent it to my aunt which has a kid that has communication problems and cannot speak.

I've looked into the contribution document and you say to call `npm install` in the root. But I saw that you have `yarn.lock` file which tells me you are using yarn to install deps. Thus `npm install` should be replaced with `yarn`. I'll open a PR, hope that it's ok :D

chiptus:
I would say that maybe npm should be changed to yarn in the whole repo, but my PR will initially be for CONTRIBUTING.md, I will do more if anyone can confirm

shayc:
Hi @chiptus, welcome!

I just wanted to say, I'm a little bit sick these days, I will hopefully feel better tomorrow and can better check everything you did! Sorry for the short message for now.

chiptus:
No worries, hope you feel better soon

shayc:
hey, i'm not sure about replacing `npm` with `yarn` just yet. the reason we used `yarn` was because it was faster (pre npm v5), now they're relatively on par perf wise.
i'll close this for now until we decide if we go back to npm (which i think is reasonable these days).

shayc:
@chiptus also feel free to drop by our Discord channel at: https://discord.gg/TEH8uxh

PR #172, opened 2018-03-09
change npm to yarn in docs

chiptus:
see isuue #171


shayc:
closing, we might be switching back to npm soon!

Issue #158, opened 2018-02-20
Testing - increasing code coverage


shayc:
We want to increase code coverage to 100%, at the moment we're below 20%, I would start with presentational components first, as these are simple to test and would mostly not change.

- [ ] `About`
- [ ] `ColorSelection`
- [ ] `FormItems`
- [ ] `FullScreenDialog`
- [ ] `Grid`
- [ ] `InputImage`
- [ ] `LoadingIcon`
- [ ] `Notifications`
- [ ] `SelectedCounter`
- [ ] `Settings` (sub components)
- [ ] `SymbolSearch`
- [ ] `WelcomeScreen`

johanzilber:
Working on:
- snapshot testing framework
- snapshot test for ```About``` component

johanzilber:
test for: 
```ColorSelection```
```FormItems```

https://github.com/cboard-org/cboard/pull/163

chiptus:
I think this would be a place to comment, that Select snapshot test is failing. Some space character...
![image](https://user-images.githubusercontent.com/1381655/37211340-455515e8-23b4-11e8-8d5d-8ffbd72cf58f.png)

chiptus:
Can I work on this? e.g on the Loading Icon?

johanzilber:
Snapshot test will fail every time the component was changed. If this was intentional change and not a bug you should override the old snapshot by the new updated one.

chiptus:
It wasn't, I didn't touch `Select`, it's just what I got when I pulled the repo

shayc:
not relevant

PR #173, opened 2018-03-09
Add Snapshots for SymbolSearch and LoadingIcon

chiptus:
Related to #158 
Please review the snapshot of `SymbolSearch`, I had to use `IntlProvider` (see https://github.com/yahoo/react-intl/issues/307) but I'm not sure about the result...


shayc:
thanks! added!

PR #211, opened 2018-04-10
Minor bug fixes and cleanup

arthurdenner:
This PR simply blacklists the app reducers that holds the login/signup status and adds the code to verify if the passwords matches with Yup.

Also, some code related to the SignUp process that is not used were removed.


shayc (commented):
thanks!

Issue #224, opened 2018-04-20
Don't show signup fields after signup request is submitted


martinbedouret:
After clicking on signup button, a message is displayed asking for email confirmation.
At that moment, signup fields should not be displayed anymore, as they are confused to the user.

![screenshot from 2018-04-20 13-06-46](https://user-images.githubusercontent.com/8397930/39062156-0a5a00d4-449d-11e8-9556-6fa6aeb81bf9.png)

arthurdenner:
@shayc, can you close this, since it was implemented in #240?

PR #240, opened 2018-05-23
Refactor - Removing unnecessary data from the store

arthurdenner:
Recently, I realized that it's not necessary to keep some states in the Redux store like we did. It's simpler and more correct, in my opinion, to keep them in component's state, since it belongs only to it.

Besides these state changes, this PR solves #224, but not removing the buttons, only disabling them.


shayc (approved):
great cleanup

PR #274, opened 2018-08-03
refactor(SymbolOutput): add container and better separation

shayc:
Deleted: old `SymbolOutput`
Refactored into: `Output` a container that uses `SymbolOutput`.

Tests: none (in the future)


src/components/Board/Output/Output.container.js
```diff
@@ -0,0 +1,155 @@
...
+        image: PropTypes.string,
+        /**
+         * Label to display
+         */
+        label: PropTypes.oneOfType([PropTypes.string, PropTypes.node])
+      })
+    )
+  };
+
+  static getDerivedStateFromProps(props, state) {
```
tinchodipalma, line 50: What about:
```
static getDerivedStateFromProps({ output, intl }, { translatedOutput }) {
```

src/components/Board/Output/SymbolOutput/BackspaceButton/BackspaceButton.js
```diff
@@ -0,0 +1,41 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { withStyles } from '@material-ui/core/styles';
+import IconButton from '@material-ui/core/IconButton';
+import BackspaceIcon from '@material-ui/icons/Backspace';
+
+const styles = {
```
tinchodipalma, line 7: Can we have a CSS file for these styles?

tinchodipalma (approved):

src/components/Board/Output/SymbolOutput/BackspaceButton/BackspaceButton.js
shayc: I wish.. last time I checked `material-ui` didn't allow to run them over without `!important`.
Let me try again..

src/components/Board/Output/Output.container.js
shayc: I think it's worse to spread the object in this case.
I like function definitions to explicitly state from where the object is coming `props` or `state`. the spreaded object to me is less readable

src/components/Board/Output/SymbolOutput/BackspaceButton/BackspaceButton.js
shayc: Yeh still the same.. so in this case I think placing it inside the JS is okay.
Maybe in the future we'll switch to css in js or something similar to avoid two ways of dealing with css..

src/components/Board/Output/Output.container.js
shayc: here for example i thought it was a good fit:
```jsx
        <Scroll {...other}>
          {symbols.map(({ image, label }, index) => (
            <div className="SymbolOutput__value" key={index}>
              <Symbol image={image} label={label} />
            </div>
          ))}
        </Scroll>
```

PR #290, opened 2018-09-07
add communicator menu options

martinbedouret:


martinbedouret:
@tinchodipalma this worked fine but it shows menu items as links (in blue and underlined), do you know how to remove that format?

tinchodipalma:
Yep @martinbedouret, I'll share with you a CSS rule to fix that.

PR #316, opened 2018-10-10
Manual Scanning: move backward by pressing left and up arrow keys

tinchodipalma:


.circleci/config.yml
```diff
@@ -75,7 +75,7 @@ jobs:
             -e SENDGRID_API_KEY \
             -e JWT_SECRET \
             cboard/cboard-bootstrap \
-            up -d --no-deps cboard && exit"
+            up -d --no-deps cboard" && exit'
```
shayc, line 78: is the `"` character by mistake?

shayc (changes requested):
seems like the `"` character is a mistake, not 100% sure

.circleci/config.yml
shayc: missed!

shayc (approved):

PR #324, opened 2018-10-20
Update zh-CN.json

hanzhaogang:
61-150 lines reviewed. "Time between two consecutive scanning highlights" cannot be totally undertand and no grantee for this translation.


shayc:
No problem. When translating symbols it is usually best (not must) to check the symbol visually since ‘square’ can mean a traffic ‘square’ or the shape ‘square’. Most symbols are pretty self evident by name alone.

Issue #114, opened 2017-11-18
Create VoiceRecorder component


shayc:
Create a `VoiceRecorder` "dumb" component to record voice from the users' microphone using Media Capture API.

Implementation should have a microphone icon button (`material-ui`), when clicked, uses the browser recording API, the recorded blob should then be bubbled up through props.

shayc:
taking!

NataliaGon:
I am going to do this component

martinbedouret:
I assume this is going to be per-pictogram , so users can add voice when a board is in edit mode, like following:

![image](https://user-images.githubusercontent.com/8397930/43083907-f1536a2e-8e6d-11e8-925b-fad0661558a3.png)

SteveALee:
This sounds really useful to AlwaysInMind too so I'm watching. Happy to help test etc.

shayc:
Checking the PR, will have a detailed comment soon..

PR #330, opened 2018-10-31
VoiceRecorder Component/Container

NataliaGon:
Work in progress, solves #114


martinbedouret:
hey, great to hear from you !

shayc (changes requested):
This is our first code review :smile:
I left some comments, some lines need clarification, some need changes to code. You can fix this in your local branch, whenever you push your changes to GitHub, this PR will automatically updated with changes.

package.json
```diff
@@ -40,18 +40,22 @@
...
     "react": "^16.5.2",
+    "react-audio-recorder": "^2.2.0",
     "react-autosuggest": "^9.4.2",
     "react-dom": "^16.5.2",
     "react-grid-layout": "^0.16.6",
     "react-helmet": "^5.2.0",
     "react-intl": "^2.7.1",
     "react-jss": "^8.6.1",
     "react-loadable": "^5.5.0",
+    "react-media-recorder": "^0.7.1",
```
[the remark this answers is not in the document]
NataliaGon: this one is  used

src/components/VoiceRecorder/VoiceRecorder.js
```diff
@@ -0,0 +1,54 @@
...
+    audioURL: '',
+    iconsColor: 'black'
+  };
+
+  render() {
+    const styles = {
+      color: this.state.iconsColor
+    };
+    return (
+      <div className="container-for-recorder">
```
shayc: We follow B.E.M. class naming convention, which is an acronym for Block, Element, Modifier.

Example:
```
/* Block */
.VoiceRecorder { ... }

/* Element */
.VoiceRecorder__button { ... }

/* Modifier */
VoiceRecorder--recording { ... }
```

By convention our `VoiceRecorder` componet (or block) will have a `.VoiceRecorder` classname. If you need to add state class when it is recording you can do so with `--` like so `.VoiceRecorder--recording`.

Child elements we denote them with `__` so for example `VoiceRecorder__button`.

You should check this [article](https://csswizardry.com/2013/01/mindbemding-getting-your-head-round-bem-syntax/).. many companies use this convention, even Google at some of their projects.

src/components/VoiceRecorder/VoiceRecorder.js
```diff
@@ -0,0 +1,54 @@
...
+export default class VoiceRecorder extends React.Component {
+  state = {
+    record: false,
+    mediaRecorder: '',
+    chunks: '',
+    audioURL: '',
+    iconsColor: 'black'
+  };
+
+  render() {
```
shayc, line 64: By convention we put `render` methods last in class, all of our components follow this, so it'll be odd for developers to see it up above.

You should place it as the last method in the class

src/components/VoiceRecorder/VoiceRecorder.js
shayc: Just rename the class name to be B.E.M. like. Don't forget to update your CSS.

src/components/VoiceRecorder/VoiceRecorder.js
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
+import Mic from '@material-ui/icons/Mic';
+import './VoiceRecorder.css';
+
+export default class VoiceRecorder extends React.Component {
+  state = {
+    record: false,
+    mediaRecorder: '',
```
shayc: We want to keep the `state` object as simple as it can be. Remember that `state` changes re-render the component and we want to avoid re-rendering when nothing changes *visually*. So I think we can move `record`, `mediaRecorder` and `chunks` from the `state` object to the class instance itself. The same way you defined `state = {}` on the class, we should move the properties to the class.

src/components/VoiceRecorder/VoiceRecorder.js
```diff
@@ -0,0 +1,54 @@
...
+      dat.onstop = () => {
+        let chunksForBlob = this.state.chunks;
+        this.setState({ chunks: '' });
+        this.setState({ audioURL: window.URL.createObjectURL(chunksForBlob) });
+      };
+    } else {
+      navigator.mediaDevices
+        .getUserMedia({ audio: true })
+        .then(stream => {
+          this.setState({ mediaRecorder: new MediaRecorder(stream) });
```
shayc: With the above comment on `state` properties, this line will become `this.mediaRecorder = new MediaRecorder(stream) });`

shayc (changes requested):

src/components/VoiceRecorder/VoiceRecorder.constants.js
```diff
@@ -0,0 +1,2 @@
+export const ADD_RECORD = 'ADD_RECORD';
```
shayc: We have a naming convention for actions: `cboard/VoiceRecorder/ADD_RECORD`, this way we can know in the debugger which component is responsible to which action

src/components/VoiceRecorder/VoiceRecorder.reducer.js
```diff
@@ -0,0 +1,18 @@
+const initialState = { audioURL: '', iconsColor: 'black' };
+
+function OnRecordClick(state = initialState, action) {
+  switch (action.type) {
+    case 'ADD_RECORD':
```
shayc: Should import those case strings like `ADD_RECORD` from the `.constants` file.
The benefit in using strings from a single source of truth is when we refactor and rename our strings, we don't need to do it on each component using them.

src/reducers.js
```diff
@@ -22,6 +23,7 @@ export default function createReducer() {
     board: boardReducer,
     communicator: communicatorReducer,
     scanner: scannerProviderReducer,
-    notification: notificationsReducer
+    notification: notificationsReducer,
+    track: AddRecord
```
shayc: `track` name is too generic, `voiceRecorder` shows better intent.
`addRecord` should be renamed to `voiceRecorderReducer`

src/components/VoiceRecorder/VoiceRecorder.reducer.js
```diff
@@ -0,0 +1,18 @@
+const initialState = { audioURL: '', iconsColor: 'black' };
+
+function OnRecordClick(state = initialState, action) {
```
shayc: reducer name should be changed to `voiceRecorderReducer`

src/reducers.js
```diff
@@ -7,6 +7,7 @@ import speechProviderReducer from './providers/SpeechProvider/SpeechProvider.red
 import boardReducer from './components/Board/Board.reducer';
 import communicatorReducer from './components/Communicator/Communicator.reducer';
 import notificationsReducer from './components/Notifications/Notifications.reducer';
+import AddRecord from './components/VoiceRecorder/VoiceRecorder.reducer';
```
shayc: should be renamed to `voiceRecorderReducer`

src/components/VoiceRecorder/VoiceRecorder.reducer.js
```diff
@@ -0,0 +1,18 @@
...
+  switch (action.type) {
+    case 'ADD_RECORD':
+      return {
+        audioURL: action.newBlob,
+        iconsColor: 'black'
+      };
+    case 'START_RECORD':
```
shayc: string should come from`VoiceRecorder.constants` file

src/components/VoiceRecorder/VoiceRecorder.js
```diff
@@ -0,0 +1,71 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import { addRecord } from './VoiceRecorder.actions';
+import { startRecord } from './VoiceRecorder.actions';
+import Mic from '@material-ui/icons/Mic';
```
shayc, line 3: by convention we import 3rd party packages like `react` or `material-ui` first, then local files we're importing, so it's easier for other developers to know what comes from the outside and what from within the app.

package.json
```diff
@@ -40,18 +40,22 @@
...
+    "react-mic": "^12.2.3",
     "react-redux": "^5.0.7",
     "react-router-dom": "^4.3.1",
     "react-scannable": "0.0.16",
     "react-share": "^2.3.1",
     "react-sizeme": "^2.5.2",
+    "react-user-media": "^0.3.0",
```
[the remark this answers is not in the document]
shayc: do we use this? do we import something from `react-user-media`?

src/components/VoiceRecorder/VoiceRecorder.actions.js
```diff
@@ -0,0 +1,15 @@
+import { ADD_RECORD } from './VoiceRecorder.constants';
+import { START_RECORD } from './VoiceRecorder.constants';
+
+export function addRecord(newBlob) {
```
shayc: should rename `newBlob` to just `blob` since the word "new" doesn't really add value to the meaning. the function `addRecord` doesn't care if it's new or old

src/components/VoiceRecorder/VoiceRecorder.js
```diff
@@ -0,0 +1,71 @@
...
+    };
+    return (
+      <div className="VoiceRecorder">
+        <Mic onClick={this.handleClick} style={styles} />
+        <audio src={TrackBlob} controls />
+      </div>
+    );
+  }
+}
+function mapStateToProps(store) {
```
shayc: rename `store` to `state`

src/components/VoiceRecorder/VoiceRecorder.js
```diff
@@ -0,0 +1,71 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import { addRecord } from './VoiceRecorder.actions';
```
shayc: move local file import to end of list

shayc (changes requested):
Looks like we're almost done 👍 just a few changes some are just code style by convention.
Also you will need to delete your `yarn.lock` file and `node_modules` and re-install with `yarn`.
Your Pull Request has a problem compiling.

When you fix that i'll start QA the feature :strong:

src/components/VoiceRecorder/VoiceRecorder.constants.js
NataliaGon: To change ADD_RECORD on cboard/VoiceRecorder/ADD_RECORD? It doesn't work. I see the same kind of names in Notifications, 
export const HIDE_NOTIFICATION = 'HIDE_NOTIFICATION';
Can you say more about it? How it should be?

shayc (approved):

PR #390, opened 2019-03-24
Moved overflow

gavinhenderson:
As you can see from the first screenshot the scroll bar appears in the middle of the screen which seems out of place.

Image with scrollbar

I solved this by moving the `overflow: auto` to the parent of the container it is currently on. 

Image with new scroll bar


martinbedouret:
Aweeesome! Thanks a lot for this!

gavinhenderson:
Hey @martinbedouret thanks for merging this!

Would it be worth making the center logo resize so that no scrolling is ever needed on this page? If thats a change you want I am happy to submit a PR for this

martinbedouret:
That would be great @gavinhenderson !
You are right , scrolling is ever needed on this page should be avoided in order to display login options at the first glance.
Waiting for your PR!

PR #392, opened 2019-03-26
WelcomeScreen never requires scrolling

gavinhenderson:
The logo on the welcome screen now adjusts its size so that scrolling is never required.

No media queries are required with this solution


martinbedouret:
Impressive! @gavinhenderson 
Don't have comments to do! Will deploy to QA now

martinbedouret:
@gavinhenderson it looks great in QA! Thanks 
 image

PR #435, opened 2019-05-21
Add action button for Board and Pictogram

martinbedouret:


martinbedouret:
@shayc need your help here:

- I want to style tooltips of the actions in order to make a bigger text 
- I want to select the proper checkbook on the create tile page depending on the action 

-

shayc:
Alright, I'll take this on Friday.
Regarding tooltip font size, it is too small cross app. I would like to make it larger for every tooltip.

PR #454, opened 2019-07-12
Removes debugger from main board component.

kidcompassion:
Hi!

This is my first contribution, and it may be useless - I just found there was a debugger on line 96 of Board.component.js that crashed my local on first run. But this probably only matters if devtools is open, so if that's a deliberate choice, feel free to delete this PR. :)


martinbedouret:
@kidcompassion thank you so much! It helps, did not realize!!

PR #456, opened 2019-07-22
Add screen for user details

martinbedouret:


shayc:
Added missing translation string at #457, you can use if needed.

PR #498, opened 2019-10-22
Cordova disk-based analytics & offline image caching

nous-:
All existing beacon analytics are written to disk as well as new events that were not being tracked: playSentence, tileClicked, appLaunch, appPause & appResume.

The Cordova `FileWriter` does not support simultaneous writes (to the same file). Subsequent writes must wait until the prior write finishes via `onwriteend`, or it will throw code 7, INVALID_STATE. The analytic logger handles this.

The offline middleware monitors redux events that add new data to the system.  It pulls tile images that reference remote content, downloads and writes them to disk, then sends tile edit commands to replace the remote url with a local one.

Additional notes:
- There is no UI to indicate offline mode progress/status
- The content-security-policy needs updates. Locally I'm using an allow-all policy, which allows me to get all the remote symbol sets and user logins operating.  Consider using an allow-all for the pilot run.
- Analytic events are written as json objects, \n delineated, to a single append only file

Open items:
- Need testing around actions that fire add_board & import_board
- Consideration if the events tracked for offline mode are the complete set
- Consideration of whether any additional telemetry is needed for the pilot
- Decision on content-security-policy stance (further work or allow-all?)


src/cordova-disk-analytics.js
```diff
@@ -0,0 +1,55 @@
...
+export const log = event => {
+  if (!isCordova()) return;
+
+  writeQueue.push(buildEvent(event));
+  ensureWrites();
+
+  if (STATE_NONE === state) {
+    state = STATE_PREPARING;
+
+    getFileWriter('analytics.txt', true)
```
shayc: Can the `analytics.txt` string be put at the top of the file as a `const`?

shayc (changes requested):

shayc:
looks good to me just one cosmetic fix

src/cordova-disk-analytics.js
nous-: Done.

shayc (approved):
:fire:

nous-:
Moving all code changes to a staging branch to do a test run before sending to master.

PR #500, opened 2019-10-22
Feat/symbol search filter

shayc:
image

This is the symbol set filter inside `SymbolSearch` component.
There is a UX issue: Anytime you filter a set, the suggestions will hide and you'll need to click the search field again to show them. I couldn't solve it in time, but I don't want to let this one go.

Please test and let me know if this is good enough for the pilot.

Includes:
* **perf:** debounce for suggestion API calls (it used to hammer external APIs on each keystroke
* **component:** `FilterBar` UI component to filter search suggestions


src/components/UI/FilterBar/FilterBar.css
```diff
@@ -0,0 +1,22 @@
+.FilterBar {
+  display: flex;
+  margin-top: 16px;
+  border-radius: 4px;
+  /* TODO: remove importants */
```
nous-: Did you intend to leave these TODOs?

nous- (approved):

src/components/UI/FilterBar/FilterBar.css
shayc: Nope, but I didn't want to stall the feature.
Depends on when we need this, I can fix it probably on Friday.

martinbedouret:
@shayc I have changed this after talking to the professionals. I have used switch to dinamically the symbol sources and fixed a couple of bugs

PR #502, opened 2019-10-23
Upload offline analytics

nous-:
Button added under Settings > Export, to upload disk analytics events. This option is only shown under Cordova.  I forked from branch from other PR.  Only review [latest commit](https://github.com/cboard-org/cboard/pull/502/commits/41bb0d9196ec11d2c8c29d1a259ecfa0ef60d89f)

image
New menu option

image
Progress indicator

image
Throw alert on error. Currently 404 as endpoint has not been updated.


martinbedouret (changes requested):
@nous- the important things to me are 
 * We should allow export to both authenticated and anonymous users 
* I did not  see a mechanism to handle repeated information 
* Data to be sent to the API must not be binary, it should be JSON 
* We should include user information in case of authenticated users

nous-:
Moving all code changes to a staging branch to do a test run before sending to master.

PR #503, opened 2019-10-24
Cordova Analytics

nous-:
All existing beacon analytics are written to disk as well as new events that were not being tracked: playSentence, tileClicked, appLaunch, appPause & appResume.

The Cordova FileWriter does not support simultaneous writes (to the same file). Subsequent writes must wait until the prior write finishes via onwriteend, or it will throw code 7, INVALID_STATE. The analytic logger handles this.

The offline middleware monitors redux events that add new data to the system. It pulls tile images that reference remote content, downloads and writes them to disk, then sends tile edit commands to replace the remote url with a local one.

Analytics upload button added under Settings > Export, to upload disk analytics events. This option is only shown under Cordova.

Additional notes:
- There is no UI to indicate offline mode progress/status
- The content-security-policy needs updates. Locally I'm using an allow-all policy, which allows me to get all the remote symbol sets and user logins operating. Consider using an allow-all for the pilot run.
- Analytic events are written as json objects, \n delineated, to a single append only file

Open items:
- Need testing around actions that fire add_board & import_board
- Consideration if the events tracked for offline mode are the complete set
- Decision on content-security-policy stance (further work or allow-all?)

image
New menu option

image
Progress indicator

image
Throw alert on error. Currently 404 as endpoint has not been updated.


src/components/Settings/Export/Export.component.js
```diff
@@ -125,6 +134,32 @@ class Export extends React.Component {
                   </div>
                 </ListItemSecondaryAction>
               </ListItem>
+
+              {isCordova() && (
+                <ListItem>
+                  <ListItemText
+                    primary={'Upload Analytics'}
+                    secondary={'Upload on-disk analytics'}
```
martinbedouret: This does not allow translation. Use messages from messages file (react-intl).

src/components/Settings/Export/Export.component.js
```diff
@@ -125,6 +134,32 @@ class Export extends React.Component {
...
+                          size={25}
+                          className="Export__ButtonContainer--spinner"
+                          thickness={7}
+                        />
+                      )}
+                      <Button
+                        disabled={this.state.uploading}
+                        onClick={this.onUploadClick}
+                      >
+                        Upload
```
martinbedouret: This does not allow translation. Use messages from messages file (react-intl).

src/components/Settings/Settings.component.js
```diff
@@ -8,13 +8,15 @@ import RecordVoiceOverIcon from '@material-ui/icons/RecordVoiceOver';
 import InfoOutlinedIcon from '@material-ui/icons/InfoOutlined';
 import CloudDownloadIcon from '@material-ui/icons/CloudDownload';
 import CloudUploadIcon from '@material-ui/icons/CloudUpload';
+import CastIcon from '@material-ui/icons/Cast';
```
martinbedouret: unused. remove

martinbedouret (changes requested):

src/components/Settings/Settings.component.js
nous-: Done

src/components/Settings/Export/Export.component.js
nous-: Done

martinbedouret:
After testing of this using a real device and having the API updates deployed on production I found a number of issues that prevent this to be safely merged:

- On application launch, a large number of edit tiles events are triggered without having a correlation with a real user action.
- No data duplication mechanism implementation observed 

[analytics_1573241137_9f3a842db15c_anything@cboard.io.txt](https://github.com/cboard-org/cboard/files/3826024/analytics_1573241137_9f3a842db15c_anything%40cboard.io.txt)

PR #558, opened 2019-11-15
Fix bug in signup-verification dialog

nathanbaleeta:
Fixed bug in signup-verification dialog


martinbedouret (changes requested):
@nathanbaleeta I would take advantage you are fixing this in order to do a full refactor for the signup component. Please check how the 'ChangePassword' component was implemented.
Please :
* create an api call in 'src/api/api.js'
* create api actions for signup start , success and failure in ‘SignUp.actions.js‘
* Update Component accordingly (see line 50 in '
changePassword.component.js'

PR #605, opened 2020-01-14
Migrate material ui 4

martinbedouret:


package.json
```diff
@@ -18,9 +18,8 @@
   "license": "GPL-3.0-only",
   "dependencies": {
     "@cospired/i18n-iso-languages": "2.0.5",
-    "@material-ui/core": "^3.9.3",
-    "@material-ui/icons": "^3.0.2",
-    "@material-ui/lab": "^3.0.0-alpha.30",
```
shayc, line 23: @martinbedouret you removed `material-ui/lab`, are you sure we're not using any of the components?
image

src/components/Settings/Speech/Speech.component.js
```diff
@@ -7,7 +7,7 @@ import ListItem from '@material-ui/core/ListItem';
 import ListItemText from '@material-ui/core/ListItemText';
 import Menu from '@material-ui/core/Menu';
 import MenuItem from '@material-ui/core/MenuItem';
-import Slider from '@material-ui/lab/Slider';
```
shayc, line 10: I see now that `Slider` is part of `material-ui`

PR #679, opened 2020-04-13
Support for ESlint

martinbedouret:


shayc:
Heads-up, `create-react-app` comes with `eslint` built in, this is PR is probably not necessary, or is it?

PR #691, opened 2020-04-28
WIP: Support quick zoom for boards 

martinbedouret:


src/components/Board/BoardZoom/BoardZoom.component.js
```diff
@@ -0,0 +1,118 @@
...
+    zoomInDisabled: false
+  };
+  onZoomInClick = () => {
+    const {
+      displaySettings,
+      updateDisplaySettings,
+      activeBoardId,
+      changeBoard
+    } = this.props;
+    if (displaySettings.uiSize === DISPLAY_SIZES[DISPLAY_SIZES.length - 1]) {
```
sylvansson, line 39: I see that `lodash` is already a dependency. You could do something like this:
```javascript
if (displaySettings.uiSize === _.last(DISPLAY_SIZES)) {
  return;
}
```

PR #697, opened 2020-05-02
Fix focus style for output bar when dark theme is enabled

### LINKED ISSUE #695: When dark theme is on, when the output bar is clicked, it highlights the bar with white color so the text below the symbols can’t be seen because the text is also in white color.
When dark theme is on, when the output bar is clicked, it highlights the bar with white color so the text below the symbols can’t be seen because the text is also in white color (screenshot in the attachment). I don’t know if users track the symbols and the text while the speech for whole sentences is reproduced, so I don’t know how big of a problem this is.
Screenshot_20200429-185724_Cboard

sylvansson:
Before and after:

image
image

Close #695

### DIFF
--- a/src/components/App/App.component.js
+++ b/src/components/App/App.component.js
@@ -2,6 +2,7 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import Helmet from 'react-helmet';
 import { Route, Switch } from 'react-router-dom';
+import classNames from 'classnames';
 
 import Activate from '../Account/Activate';
 import ChangePassword from '../Account/ChangePassword';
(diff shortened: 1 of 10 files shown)


sylvansson:
@martinbedouret I revised my PR. Instead of defining a new class with a `--dark` suffix (see 0d7cb71), I decided to take @shayc's approach in 01e78cf and add the `is-dark` class to the `App` component when dark mode is enabled. 

This saves us the trouble of passing the `darkThemeActive` prop around, and of having to add a class to each component that needs special styling in dark mode.

There are three components that still use the other approach:

- `Board`;
- `Navbar`;
- `SymbolOutput`.

I can update them in this PR if you'd like.

martinbedouret:
@sylvansson I agree with you and I think your approach is better.
It would be great if you can work into your comments about components using the other approach as well.
Thank you !

sylvansson:
Hi @martinbedouret, just to be sure, you're saying you want me to update the remaining three components in this PR?

martinbedouret:
Yes @sylvansson , you are right.

sylvansson:
Great, I'll make the changes tonight.

martinbedouret (approved):
@sylvansson this is improving the code definitely. You simplified it using this approach. Thank you !

PR #698, opened 2020-05-02
Ensure tile and label are on the same page in PDF export

### LINKED ISSUE #692: When the option Above for Label position is selected, the exported board will have the label positioned below from the second page of the PDF
: when the option Above for Label position is selected, the exported board will have the label positioned above the symbol which is fine, but from the second page of the PDF, one extra row of labels is shown (although the symbols are on the next page). I am sending a PDF in the attachments so you can see the situation.

sylvansson:
This PR ensures that a tile and its label are always on the same page when exporting to PDF, by adding a page break every 4 rows.

Close #692


martinbedouret (approved):
It looks great. An elegant solution.

PR #705, opened 2020-05-29
fix: color scheme for dark mode

### LINKED ISSUE #694: When dark theme is on, the links that appear in the app are not visible because of the blue color

### LINKED ISSUE #704: Enhancement for the output bar in dark mode
Currently we have:
image

I'm suggesting maybe:
image

The problem I see with current theme, navbar is too bright (for a dark theme), some symbols were mostly designed for a white background and are less legible this way. So to be on the safe side we can have the output bar stay white..

shayc:
fixes dark color scheme with symbols legibility.
this PR implements what is suggested in the screenshots of #704 and #694

### DIFF
--- a/src/components/Board/Board.css
+++ b/src/components/Board/Board.css
@@ -63,7 +63,7 @@
 }
 
 .is-dark .Board__tiles {
-  background: #303030;
+  background: #444;
 }
 
 .Board__tiles .Grid {
--- a/src/components/Board/Navbar/Navbar.css
+++ b/src/components/Board/Navbar/Navbar.css
@@ -7,11 +7,6 @@
   color: #fff;
 }
 
-.is-dark .Navbar {
-  color: #000;
-  background: #fafafa;
-}
-
 .Navbar__title {
   display: flex;
   order: 2;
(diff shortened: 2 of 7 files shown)


martinbedouret (changes requested):
@shayc  fix for links looks great !
For the output bar I would rather take a different strategy: 

1. Replace the inline CSS of the SVG images by a class name 
2. Control the stroke color of the SVGs from CSS 

That will give us a lot of flexibility for the future in case we want to implement more themes or change the contrast of symbols for people with visual impairments. Here an example of the result:
image

martinbedouret:
https://stackoverflow.com/questions/37940282/changing-the-stroke-color-for-an-inline-svg

shayc:
@martinbedouret your approach is better but we have a few obstacles: 

1) At the moment we use `` to render SVG, this limits us technically from styling SVG with CSS. We can refactor the code to render with `<svg />` tags instead but other file formats (jpg/png) still need `` tags to render and we wont be able to style them for dark theme, this will cause some of them to not be legible.

Example: 
image

2) It will probably require some kind of modification to the SVG files themselves. I'm not sure you can easily target just the borders of all SVGs in a generic way, unless they were made in a certain way, we need to investigate this with a small POC.

I thought maybe inverting the colors with CSS can be a solution, but it looks weird.

image

martinbedouret (approved):

PR #709, opened 2020-06-09
Add client based Analytics using Google analytics reports 

martinbedouret:
This PR adds a new analytics page for users to check their analytics information based on the google analytics client id value.
This approach has some advantages and disadvantages:

### **Pros:**

- Data gathering and display completely anonymous 
- No need for tracking, log or store custom data in the Cboard back-end 
- Re-use existing GA code we already have in Cboard client.

### **Cons**

- Data requests are based on the google analytics **client id** value, which is a unique, randomly generated string that gets stored in the browsers cookies. Hence, data will be different for every device and browser.

We need to discuss if reports should be generated based on the google analytics **user id** instead. See https://developers.google.com/analytics/devguides/collection/analyticsjs/cookies-user-id

### **Note for reviewers:**
There are many files on this PR, so my suggestion is you guys to focus just on two files:

1. ccboard\cboard\src\components\Analytics\Analytics.component.js
2. ccboard\cboard\src\components\Analytics\Analytics.container.js

Some images:

image

image

image

image


sylvansson:
Hi @martinbedouret, I'll review tomorrow or Saturday. The screenshots look great!

src/components/Analytics/Analytics.container.js
```diff
@@ -0,0 +1,384 @@
...
+    } catch (err) {
+      this.setState({ isFetching: false });
+      showNotification(intl.formatMessage(messages.loadingError));
+      console.log(err.message);
+    }
+  }
+
+  getSymbolSources() {
+    const { boards } = this.props;
+    const images = boards
```
sylvansson, line 88: Nitpick: You can probably use [`flat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat) instead of `reduce` here.

src/components/Analytics/Analytics.container.js
```diff
@@ -0,0 +1,384 @@
...
+      this.timerId = setTimeout(() => {
+        if (
+          typeof window.ga !== 'undefined' &&
+          typeof window.ga.getAll === 'function' &&
+          typeof window.ga.getAll()[0] !== 'undefined' &&
+          typeof window.ga.getAll()[0].get('clientId') !== 'undefined'
+        ) {
+          resolve(window.ga.getAll()[0].get('clientId'));
+        } else {
+          reject(new Error({ message: 'Google analytics client idnot found' }));
```
sylvansson: Missing space between "id" and "not".

sylvansson (approved):
I have a few nitpicks (see my comments) but the code looks good to me otherwise. I wasn't able to sign up locally so I couldn't actually test the changes.

It's hard to say whether using the client id or the user id is preferable. I would say it depends on several factors like:

- Who will be using the analytics page? An individual user, a therapist, etc.
- Do they _actually_ use the app on multiple devices and browsers?
- Is the client id deterministic? In other words, if I clear my cookies, do I get a different client id and therefore lose my analytics?

I'm new to this field so I might be making false assumptions, but it seems to me that an individual user is more likely to use a single dedicated device, in which case either id might be fine. But a therapist might use the app with different patients across multiple devices, in which case aggregating by user id is probably not desirable.

We could give the user the pros and cons and let them choose. In any case, I think the tracking itself should be opt-in. 

(I also _believe_ that if we use the user id, we need a mechanism for users within the EU to request the deletion of their data to be GDPR compliant.)

src/components/Analytics/Analytics.component.js
```diff
@@ -0,0 +1,242 @@
...
+    }
+    this.setState({
+      openDetailsDialog: true
+    })
+  };
+
+  getDates = range => {
+    const days = [];
+    const dateEnd = moment();
+    const dateStart = moment().add(-range, 'days');
```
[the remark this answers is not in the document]
martinbedouret: absolutely right

src/components/Analytics/Analytics.container.js
```diff
@@ -0,0 +1,388 @@
...
+        }
+      });
+      return all;
+    }, {});
+    let total = 0;
+    Object.entries(summary).forEach(([key, value]) => {
+      total = total + value;
+    });
+    let summaryData = [];
+    Object.entries(summary).forEach(([key, value]) => {
```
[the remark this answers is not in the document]
martinbedouret: done

src/components/Analytics/Analytics.container.js
martinbedouret: thank you

martinbedouret:
Awesome comments @sylvansson , appreciate it !!

Regarding the **client id versus user id** implementation, the response was in the Google analytics documentation:
https://support.google.com/analytics/answer/6205850#reporting
image

We are not allowed to get reports using the user id, so no other option than the **client id.**

@shayc could you let me know what is failing on synk check ? I don't have access to it.

PR #721, opened 2020-06-21
chore: upgrade pdfmake from 0.1.39 to 0.1.65

shayc:
`pdfmake` was upgraded

Some of the changes in the lib:
image


shayc:
It appears for me that on Firefox (in production) the pdf/print feature is missing some symbols. The behavior is still the same after this update (no regression). I'll open an issue.

martinbedouret (approved):

PR #739, opened 2020-07-03
Generate just an OBF file when exporting a single board

### LINKED ISSUE #653: Allow exporting just a single board using the obf file format

sylvansson:
This PR changes the behaviour when exporting a single board in the Open Board Format to generate a single OBF file rather than an OBZ archive. I tested this in my browser and everything seems fine, but I'll do another round of QA tomorrow evening. I suggest someone else test this too.

A few notes:

- Images are embedded in the file with the `data` property. It would be useful to embed sounds as well, but it seemed out of scope so I simply added a TODO.
- OBF files are pretty printed to make them easier to read and edit manually.
- Exports in the Open Board Format are now prefixed with the date, and the time uses the 24-hour clock. This makes exports easier to sort chronologically.

@martinbedouret @shayc Have we considered or tried using TypeScript before? I started using it for this task but it got slightly out of hand, so I undid the changes. It seems we could use the type safety.

Close #653


martinbedouret (commented):
I have been testing it and worked great. Still need to do testing using symbols from APIs and do some code review.

src/components/Settings/Export/Export.helpers.js
```diff
@@ -409,7 +435,42 @@ const getDisplaySettings = () => {
...
+  const { obf } = await boardToOBF({ [board.id]: board }, board, intl, {
+    embed: true
+  });
+  const content = new Blob([JSON.stringify(obf, null, 2)], {
+    type: 'application/json'
+  });
+
+  if (content) {
+    // TODO: Remove illegal characters from the board name.
+    const prefix = moment().format('YYYY-MM-DD_HH-mm-ss-') + board.name + ' ';
```
martinbedouret: I like the naming convention, anyway we should also modify it for the other adapters (pdf and cboard ).

martinbedouret (changes requested):
Great job @sylvansson ! My only comment is to use the same naming convention you are using for obf's, on the other two types (pdf and cboard ).

martinbedouret (approved):

PR #740, opened 2020-07-04
Do not delete board when removing a folder tile

### LINKED ISSUE #629: Unexpected deletion of board upon folder tile removal
Since the introduction of existing board linking, removing a folder tile linking to the board will result in the deletion of the linked board. While that made sense before when linking to existing boards was not possible, now this deletion results in broken links for the other boards. Here is an illustration of what is happening:

Board_Deletion

1. Board B is created by adding an empty folder in Board A. An additional link to Board B is done by adding a folder in Board C.
2. The folder linking to Board B is removed from Board A. As a consequence, Board B is deleted, and the link from Board C is no longer valid.
2. The folder linking to Board B is removed from Board C. As a consequence, Board B is deleted, and the link from Board A is no longer valid.

Ultimately, a viable solution would be to leave Board B intact if either the folder in Board A or Board C is removed. The removal of Board B could be done in the build boards menu exclusively.

sylvansson:
Close #629

### DIFF
--- a/src/components/Board/Board.container.js
+++ b/src/components/Board/Board.container.js
@@ -40,15 +40,13 @@ import {
   updateApiObjects,
   updateApiObjectsNoChild,
   getApiObjects,
-  deleteApiBoard,
   downloadImages,
   updateApiBoard
 } from './Board.actions';
 import {
   upsertCommunicator,
   changeCommunicator,
-  addBoardCommunicator,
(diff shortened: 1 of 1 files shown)


martinbedouret (approved):

remusonline:
Excellent, cant wait to try the version with the fix. Thanks everyone! 😁

PR #760, opened 2020-07-26
Store images as data URIs when importing from OBF/OBZ

### LINKED ISSUE #756: Keep embedded images when importing an OBF board that contains embedded images
Currently, when a user imports an OBF board that contains embedded images, we save those images in the cloud and we replace the tile image by a reference to the file in the cloud. That affects to the offline users as the cloud images won't be displayed.
The initial solution would be to simply keep embedded images in such format `data:image/...` but I am open to any other suggestion.

sylvansson:
This PR changes the behaviour when importing boards in the Open Board Format to no longer upload files and instead store them as data URIs. This allows users to see images even when they're offline. 

A few more things that are out of scope, but still included in this PR:

- I noticed an issue with the export to OBZ. From what I understand, the image ids in `manifest.json` need to match the ones in the OBF files. This is fixed in 78e32a7.
- The content type wasn't set properly after #739. This is fixed in 919d3a6.

I did some QA but I encourage everyone to give it a try, especially with Cordova. There are _many_ scenarios to test.

Close #756

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "keycode": "^2.2.0",
     "lodash": "^4.17.16",
     "matx": "1.0.0",
+    "mime-types": "^2.1.27",
     "moment": "2.27.0",
     "opencollective": "^1.0.3",
     "pdfmake": "^0.1.66",
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
     "matx": "1.0.0",
     "mime-types": "^2.1.27",
     "moment": "2.27.0",
+    "mongoose": "^5.9.25",
     "opencollective": "^1.0.3",
     "pdfmake": "^0.1.66",
     "prop-types": "^15.7.2",
--- a/src/components/Settings/Export/Export.helpers.js
+++ b/src/components/Settings/Export/Export.helpers.js
@@ -88,7 +88,7 @@ function getBase64Image(base64Str = '') {
   };
 }
 
-async function getDataUri(url) {
+export async function getDataUri(url) {
   try {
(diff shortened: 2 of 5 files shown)


martinbedouret (commented):
@sylvansson update looks good. Anyway, I still feel we need to evaluate if this feature is really a good thing or not.
The advantage I see is that the imported board can be used 100% offline. On the other side, we introduced a disadvantage that is that we are saving images into the DB and increasing size for the board, using resources less efficiently. Sorry about my late thought. What do you think?

sylvansson:
I would have to take a closer look at how we handle remote images. For example, if I open a board on my phone while I have an internet connection, then close and reopen the app when I no longer have a connection, do I still have access to the images? In other words, are images cached somehow?

But I personally think it makes more sense to store everything as a data URI, because it ensures that the user can access an image even if:

- They have no internet connection, a slow connection, or very limited mobile data.
- The image becomes unavailable, either temporarily or permanently.

Perhaps we could give users the option of converting remote images to data URIs, and vice versa. This way they could decide between using more space on their device versus needing internet access.

martinbedouret:
@sylvansson after some research on how other tools are handling the same thing, I have decided to keep the cloud storage for importing boards and to store the data url just if the cloud connection fails. Large sizes for the board files are discouraged, as other problems will arise, such as the request quota on browser storage (I got this doing testing of your PR):
image
Hence, I have reverted the update for importing but I kept your fixes for exporting, good catch !

sylvansson:
Oops, thank you for doing some research and QA.

PR #767, opened 2020-08-04
Avoid recursive function when converting images to data URI

### LINKED ISSUE #763: BMP Images are lost when exporting a single board to OBF format
1. Import the attached board, it contains cloud images on the major formats.
2. Export the previous board as a single obf format.
3. Check the content of the obf file 
4. BMP is lost 

[images formats board.zip](https://github.com/cboard-org/cboard/files/5013146/images.formats.board.zip)

sylvansson:
Large images stored remotely currently cannot be exported because the function to convert them to a data URI is recursive, which causes us to exceed the maximum call stack size.

This PR replaces the function call with a reduce operation to avoid getting a `RangeError`.

Close #763

### DIFF
--- a/src/components/Settings/Export/Export.helpers.js
+++ b/src/components/Settings/Export/Export.helpers.js
@@ -98,7 +98,10 @@ export async function getDataUri(url) {
 
     // Convert the array buffer to a Base64-encoded string.
     const encodedImage = btoa(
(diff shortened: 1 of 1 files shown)


sylvansson:
@martinbedouret This was tested locally, and I was able to export the BMP image successfully.

martinbedouret (approved):

PR #793, opened 2020-09-27
Include subfolders when exporting single board to Cboard format

### LINKED ISSUE #773: Imported board with folders (Cboard format) has "missed folders"
After the board that has folders is exported from Cboard and then imported back into Cboard (different user), the folders appear as missing (message: We are sorry but we have missed this folder / board. We recommend you to create it again.).

However, if the board and the folders are exported and imported one by one, they will connect correctly and will not be missed.

sylvansson:
Close #773


sylvansson:
@martinbedouret Would you have time to review this PR?

sylvansson:
No worries @martinbedouret. It's just good to have you back :-)

martinbedouret (approved):

PR #813, opened 2020-12-07
Edited positions of tiles are not remembered after logout

### LINKED ISSUE #774: Edited positions of tiles are not remembered after logout
After creating a new board, then folder, and in the folder 12 symbols, the board was successfully exported as PDF board:
Screenshot_20200804-111335_Cboard
Screenshot_20200804-113053_Drive

Then the positions of tiles were edited and after that the PDF board was also successfully exported:
Screenshot_20200804-113059_Drive

But, when this folder is exported in Cboard format and imported into Cboard with different user, the positions of tiles were initial. After coming back to the user that created the board, the positions of tiles were not remembered.

martinbedouret:


sylvansson:
Hi @martinbedouret, I'll do my best to review your two PRs tonight.

src/components/Board/Board.container.js
```diff
@@ -718,6 +718,32 @@ export class BoardContainer extends Component {
     }
   };
 
+  handleLayoutChange = (currentLayout, layouts) => {
+    const { updateBoard, replaceBoard, board } = this.props;
+    currentLayout.sort((a, b) => {
```
sylvansson, line 723: I believe you can do the same thing with `_.sortBy(layouts, ['y', 'x']);`. It won't modify the array in place though.

sylvansson (approved):
Unable to test at the moment, but I assume that you did. LGTM.

PR #814, opened 2020-12-08
Shared boards have some titles of symbols translated back to English

### LINKED ISSUE #794: Shared boards have some titles of symbols translated back to English
After creating a new board with Croatian titles of symbols, I shared it via email so that I can open it and edit in the browser. After opening that board in the browser, I noticed that some of the symbols titles got back to their original English titles. After testing adding new symbols from different symbol repositories, I noticed that only titles taken from Mulberry set of symbols get back to their original English version (it doesn't matter whether the browser is on English or Croatian). 
But, if you make a small change in the title, the title remains on Croatian. So this only applies to titles taken directly from the Mulberry symbol set.

martinbedouret:


src/components/Board/Board.container.js
```diff
@@ -1093,11 +1099,11 @@ export class BoardContainer extends Component {
 
   handleCopyRemoteBoard = async () => {
     const { intl, showNotification } = this.props;
-    await this.createBoarsRecursively(this.state.copyPublicBoard);
+    await this.createBoardsRecursively(this.state.copyPublicBoard);
     showNotification(intl.formatMessage(messages.boardCopiedSuccessfully));
   };
 
-  async createBoarsRecursively(board) {
```
sylvansson, line 1100: image

sylvansson (approved):
I'm not able to test locally at the moment, but I'll give it a try once deployed. LGTM.

src/components/Board/Board.container.js
martinbedouret: Hehehe

PR #826, opened 2021-02-02
Intuitive board share

### LINKED ISSUE #825: Refactor share board dialog
We want to update the share board dialog, in order to make it more intuitive. 
This is the current dialog: 

image

We want to do something like this: 

image

If the board is already a public board, the dialog should display: 

image

tomivm:
issue  #825
1.Can I log in to my local server?
I modified 4 files, but I only edit BoardShare Files. it is because I did a "git add ." in cboard directory? How I can fix this? 
What's better? "now is working" or "wat's better" commits.
it was a better option to create a new component ButtonPublish, and with a prop "isPublic" return de correct button?


martinbedouret:
@tomivm I send you some initial feedback 

- Never commit lock files (package-lock.json, yarn.lock).
- Every time you have added a new string into a *.messages.js file, please add an entry into this file: https://github.com/cboard-org/cboard/blob/master/src/translations/src/cboard.json
- Please also update the title of the dialog to be 'share your board '

martinbedouret (changes requested):

tomivm:
Hello! It's ready for a new review. Now I'm havings problems with the default messages of boardshare.messages.js . The component inspector of react shows the correct value. But its no reflecting in the modal.
imagen

martinbedouret (approved):

PR #855, opened 2021-03-03
Support Tetum language from Timor - Leste 

### LINKED ISSUE #848: Support Tetum language (from Timor -Leste)

martinbedouret:


adelinodst:
Issue when upload file to app in offline mode sounds cannot heard but build in language sound is ok
is there a solution?
Thanks

PR #860, opened 2021-03-09
update new content message adding kinds of notifications

### LINKED ISSUE #835: update the new version message
We want to update the new version message. This is how it looks now: 

image

We want something like this: 

image

Use the following component: https://material-ui.com/components/snackbars/#customized-snackbars

tomivm:


sylvansson (commented):
This looks OK to me but I'm not a frontend developer so let's see what @martinbedouret thinks :-) 

I would suggest manually running Prettier on the files that you modified, because there are some minor styling issues like too much indentation. It usually runs automatically in a pre-commit hook but it might not be configured correctly on your machine.

src/components/Notifications/Notifications.reducer.js
```diff
@@ -3,14 +3,15 @@ import {
   HIDE_NOTIFICATION
 } from './Notifications.constants';
 
-const initialState = { message: '', open: false };
+const initialState = { message: '', open: false , kind : null };
```
martinbedouret: use `undefined`, In JavaScript, undefined is a type, whereas null an object.

martinbedouret (changes requested):

martinbedouret (approved):

PR #880, opened 2021-04-09
Implement copy button

### LINKED ISSUE #861: Make current phrase shareable
I would like to be able to  share the current phrase,  for example to send it via email or copy it into a word processor.

I could imagine multiple ways to support this:
- copy to clipboard by  ctrl+c
- add a copy to clipboard symbol next to "clear" and "delete"
- support dragging of the phrase to some other programm
- add a "share" button to select further options (at best user configurable providers)

aushaag:
Create new copy button in the symbol output (disabled by default). When users click the copy button, copy the labels of the selected symbols to the clipboard:

image

Disable CopyButton by default, and give users the option to enable it in Settings/Navigation/ShowCopyButton:

image


src/components/Board/Output/Output.container.js
```diff
@@ -179,6 +181,16 @@ export class OutputContainer extends Component {
     this.clearOutput();
   };
 
+  handleCopyClick = () => {
+    const { intl, showNotification } = this.props;
+    let labels = [];
+    this.props.output.map(symbol => {
```
martinbedouret: remove function curly brackets if map does contain a single line

martinbedouret (changes requested):
@aushaag Terrific job!! Congrats! 
My only comment is that you don't need to worry about the translation files. You only need to add any new string to be translated into the `cboard\src\translations\src\cboard.json` file.  before the production release, an automation script will take care of the translations.

src/components/Board/Output/Output.container.js
aushaag: Got it, I think I fixed this with the most recent commit

martinbedouret (approved):

src/components/Board/Output/Output.container.js
```diff
@@ -179,6 +181,14 @@ export class OutputContainer extends Component {
     this.clearOutput();
   };
 
+  handleCopyClick = () => {
+    const { intl, showNotification } = this.props;
+    let labels = [];
```
sylvansson, line 186: I know this was merged already, but I have a very small comment. 

It would be simpler to assign the value when declaring the variable, and it's usually preferable to use `forEach` over `map` when we don't care about the function's return value.

```javascript
```

sylvansson (commented):
Thanks for the great first contribution @aushaag :-)

src/components/Board/Output/Output.container.js
aushaag: That makes sense. Just submitted a request for this update, thanks!

PR #892, opened 2021-04-19
Iss886 update tts

### LINKED ISSUE #886: Update TTS functions to support the Electron platform
We have to update the cordova tts engine related commands in order to avoid errors on the electron platform. Please check the following file: 
`board\src\providers\SpeechProvider\tts.js`

We have to update the following functions: 

- `setTtsEngine(ttsEngineName)`
- `getTtsEngines()`
- `getTtsEngines()`

RodriSanchez1:
#886 For fix the issue, i added two new functions: `isAndroid()` and `isElectron()`  in `cordova-util.js` which check if you are in Cordova-Android-Platform or Cordova-Electron-Platform. 
In `tts.js` file, all the ttsEngines are asociated to android.
In Language component only render the select engine if you are in android platform


martinbedouret:
@RodriSanchez1 Changes look great, please update `cboard\src\providers\SpeechProvider\SpeechProvider.container.js` on line 43 as well.

src/cordova-util.js
```diff
@@ -1,5 +1,11 @@
 export const isCordova = () => !!window.cordova;
 
+export const isAndroid = () =>
+  isCordova() && window.cordova.platformId === 'android' ? true : false;
```
sylvansson: This is a very small detail, but you don't need to use a ternary (conditional) operator here. The condition is already a boolean so you can leave out `? true : false`.

sylvansson (commented):
I don't think we should be deleting the lockfile.

martinbedouret (approved):

PR #900, opened 2021-04-22
implement redux logic to board 'root button'

### LINKED ISSUE #630: Context aware back button that looks like it should take a user to the top (first) board
I am just wondering if the context aware button that looks like a 'go to the first board'  should take the user back to the home board (top board) rather than just one back if one has three layers.  It has a line and arrow indicating the end of something or the beginning.  I have made some PowerPoint slides to illustrate the issue as this arrow could be helpful if there are three layers when linking folders. 
[Keeping links down to three layers.pptx](https://github.com/cboard-org/cboard/files/4203619/Keeping.links.down.to.three.layers.pptx)

tomivm:
This PR creates a reducer and dispatch an action when the 'To root button' is pressed.
is this the correct good way to dispatch an action?

It would better to get a register for later add a button to return to the previous board after press the 'to root button'? has sense?


src/components/Board/Board.container.js
```diff
@@ -1120,6 +1125,14 @@ export class BoardContainer extends Component {
     }
   }
 
+  onRequestToRootBoard() {
```
sylvansson, line 1128: I don't _think_ we want to return to the default ("root") board here, we probably want to return to the top board that was last loaded. So for example, if I open a board B and I click multiple folders, the button should take me back to board B, not the default board that's in `/board/root`.

@eadraffan @martinbedouret

tomivm: I agree :) I will correct that

src/components/Board/Board.container.js
```diff
@@ -1120,6 +1125,14 @@ export class BoardContainer extends Component {
...
+    const { navHistory,toRootBoard } = this.props;
+    toRootBoard();
+    if (this.props.navHistory.length >= 2) {
+        this.props.history.replace(navHistory[0]);
```
martinbedouret, line 1132: this must be done by the reducer.

martinbedouret (changes requested):

PR #907, opened 2021-05-08
Add try/catch block to share sentence function #901

### LINKED ISSUE #901: Share sentence function needs a try/catch
`Output.container.js`
Function: `handleCopyClick`

1) Need to wrap the code with a try/catch block since not all browser support `navigator.clipboard.writeText` and besides we want to be a little more defensive in our coding style.
2) Needs to add a new translation message for the "Failed to copy" string

Example:
```js
    try {
      await navigator.clipboard.writeText(labels.join(' '));
      showNotification(intl.formatMessage(messages.copyMessage));
    } catch (err) {
      showNotification(
        `${intl.formatMessage(messages.failedToCopy)}: ${err.message}`
      );
    }
```

katerinadimatou:


src/components/Board/Output/Output.container.js
```diff
@@ -184,8 +184,14 @@ export class OutputContainer extends Component {
...
     const { intl, showNotification } = this.props;
     const labels = this.props.output.map(symbol => symbol.label);
-    navigator.clipboard.writeText(labels.join(' '));
-    showNotification(intl.formatMessage(messages.copyMessage));
+    try {
+      navigator.clipboard.writeText(labels.join(' '));
+      showNotification(intl.formatMessage(messages.copyMessage));
+    } catch (err) {
+      showNotification(
+        `${intl.formatMessage(messages.failedToCopy)}: ${err.message}`
```
martinbedouret: exception `error.message ` should not be displayed to the user. Log it to the console.

martinbedouret (changes requested):
congratulations on your first pr! Great job @katerinadimatou @mariaspastef

src/components/Board/Output/Output.container.js
katerinadimatou: Is it okay now? Please tell us if we have to correct anything else

tomivm: if you catch an error, you should showNotification 'failed to copy' to the user.  And only console log `err.message`. It Is a message for dev.

martinbedouret (approved):

src/components/Board/Output/Output.container.js
```diff
@@ -184,8 +184,13 @@ export class OutputContainer extends Component {
   handleCopyClick = () => {
     const { intl, showNotification } = this.props;
     const labels = this.props.output.map(symbol => symbol.label);
-    navigator.clipboard.writeText(labels.join(' '));
-    showNotification(intl.formatMessage(messages.copyMessage));
+    try {
+      navigator.clipboard.writeText(labels.join(' '));
```
sylvansson, line 188: @martinbedouret or @shayc could confirm, but I don't think we should have removed the `await` keyword in d0dc13b6c33f2d7b99c2c83f3ac92ec346202c9f. `writeText` returns a Promise, so if we don't await it, we can't catch the error. This means we will always get a success message and never an error message, even if we fail to write to the clipboard.

@katerinadimatou Instead of removing `await`, you should be able to fix the compilation error by adding `async` to the function definition, like this:

```javascript
handleCopyClick = async () => { ... }
```

sylvansson: Obviously it's too late to fix it in this pull request, you would have to create a new one :-)

shayc: @sylvansson you are correct. the `writeText` method returns a Promise and should be awaited: [example](https://web.dev/async-clipboard/#writetext())

@katerinadimatou this will require a new PR (small change as @sylvansson suggested)

Overall good first PR! Thanks @katerinadimatou :)

PR #910, opened 2021-05-09
Change handleCopyClickfunction to async and include await

katerinadimatou:
@sylvansson I think we corrected it as you said! 
@martinbedouret & @shayc if you prefer the function like this merge these changes too or else we can leave it as it was before.


src/components/Board/Output/Output.container.js
```diff
@@ -181,14 +181,14 @@ export class OutputContainer extends Component {
...
-  handleCopyClick = () => {
+  handleCopyClick = async() => {
     const { intl, showNotification } = this.props;
     const labels = this.props.output.map(symbol => symbol.label);
     try {
       navigator.clipboard.writeText(labels.join(' '));
       showNotification(intl.formatMessage(messages.copyMessage));
     } catch (err) {
-      showNotification(intl.formatMessage(messages.failedToCopy));
+      await showNotification(intl.formatMessage(messages.failedToCopy));
```
sylvansson: @katerinadimatou Just to clarify, we want to await `writeText`, not `showNotification`. To help you understand why it's needed, you can run this in your console:

```javascript
try {
  // Try with and without `await` here.
  Promise.reject(new Error('Something went wrong!'));
  console.log('Success');
} catch (err) {
  console.log(err.message);
}
```

You will see a success message even though the Promise was rejected, which is incorrect. That's because a rejected Promise doesn't throw an error unless you await it, so we can't catch the error. Promises can be tricky to use, so if you're not familiar with them I would suggest reading tutorials like this one:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises 

Feel free to also reach out to me on Discord if needed.

sylvansson (changes requested):

src/components/Board/Output/Output.container.js
katerinadimatou: I believe we are ready now! @sylvansson @shayc

sylvansson (approved):
Looks good to me but let's wait for @shayc or @martinbedouret to approve too. Thank you for fixing this!

katerinadimatou:
> Looks good to me but let's wait for @shayc or @martinbedouret to approve too. Thank you for fixing this!
Our pleasure! Sorry, it was a stupid mistake!

shayc (approved):

PR #926, opened 2021-05-26
Fixes for the share function on Android

### LINKED ISSUE #768: Sharing board via Copy link on Android is not intuitive
Copy link option for sharing boards on Android application creates a link such as:  file:///android_asset/www/index.html#/board/root

but the user will not know what to do with it because it is not a classic URL web address. Maybe it would be better if this link is also an URL like the one that is created when sharing with WhatsApp? That way the user can share the link via other message applications or something.

martinbedouret:


tomivm (commented):
I tested this using BrowserStack and a galaxy s10 with android V9.
it works fine!
image
the pasted link:
image

tomivm (approved):

PR #934, opened 2021-06-08
Fixes for the read / write asynchronous functions to files in cordova

### LINKED ISSUE #940: Cordova - Electron - Image doesnt load if user is not logged in
When you try to load an image to create a new folder or tile in electron app, its never showed. That problem only appears when the user is not logged in or doesnt have internet conection. 
 The bug is in the locally save method.

martinbedouret:


RodriSanchez1 (approved):
Its okay, good changes.

PR #941, opened 2021-06-11
Electron input image fixed

### LINKED ISSUE #940: Cordova - Electron - Image doesnt load if user is not logged in

RodriSanchez1:
#940 
The load function in reader object was changed, to other one that have the same functionality and works with Electron.


src/components/UI/InputImage/InputImage.component.js
```diff
@@ -34,13 +35,9 @@ class InputImage extends PureComponent {
...
-    reader.addEventListener(
-      'load',
-      () => {
-        this.props.onChange(reader.result);
-      },
-      false
-    );
+    reader.onload = () => {
+      this.props.onChange(reader.result);
+    };
```
martinbedouret, line 40: @RodriSanchez1 Did you test this both for android and electron?

RodriSanchez1: Yes, i did it in both platforms!

PR #952, opened 2021-06-17
Implement online voices from Azure 

### LINKED ISSUE #788: Support for online voices from any tts vendor (Azure, AWS, Google cloud, etc)

### LINKED ISSUE #959: Setting component doesn't work properly
When the setting component is rendered. It does twice.

This is an example in the branch feature/azure-tts
https://drive.google.com/file/d/15m5jlFQ9U7tnCZLkIpkLsMpa3sOVmM7o/view

martinbedouret:


src/providers/SpeechProvider/SpeechProvider.actions.js
```diff
@@ -124,15 +124,40 @@ export function changeRate(rate) {
...
+              voice.voiceURI = voiceURI;
+              voice.voiceSource = 'local';
+            } else if (ShortName) {
+              voice.voiceURI = ShortName;
+              voice.voiceSource = 'cloud';
+            }
+            if (name) {
+              voice.name = name;
+            } else if (DisplayName) {
+              voice.name = `${DisplayName}  (${voice.lang}) - ${Gender}`;
```
sylvansson: Should we be using the messages in `SpeechProvider.messages.js` here?

src/constants.js
```diff
@@ -12,6 +12,12 @@ export const TAWASOL_BASE_PATH_API =
   'https://madaportal.org/tawasol/wp-json/custom/v2/dea44ade76/';
 export const TAWASOL_BASE_IMAGE_URL =
   'https://www.madaportal.org/tawasol/wp-content/uploads/images/';
+export const AZURE_VOICES_BASE_PATH_API =
+  'https://eastus.tts.speech.microsoft.com/cognitiveservices/voices/';
```
sylvansson: Are we only getting American voices if we list voices for the East US region? Should we be using `AZURE_SPEECH_SERVICE_REGION` here?

src/providers/SpeechProvider/SpeechProvider.actions.js
```diff
@@ -124,15 +124,40 @@ export function changeRate(rate) {
...
+          ({
+            voiceURI,
+            lang,
+            name,
+            Locale,
+            ShortName,
+            DisplayName,
+            Gender
+          }) => {
+            let voice = {};
```
sylvansson: It might be easier to do this transformation in `getVoices` (or maybe even `getAzureVoices`) since we already know if the voice is provided by Azure or not, and we wouldn't have to check if different properties are defined or not. We could also write some JSDoc for `getVoices` and have a custom `Voice` type:

```js
/**
 * @typedef {Object} Voice
 *
 * @description A voice that can be used for text-to-speech.
 *
 * @property {string} lang
 * The voice's language code. For example, "en-US".
 *
 * @property {string} voiceURI
 * The voice's identifier. For example, "en-US-AriaNeural".
 *
 * @property {('local'|'cloud')} voiceSource
 * Whether the voice is cloud-based and requires an Internet connection,
 * or is provided by the platform.
 *
 * @property {string} name
 * The voice's display name. For example, "Aria (en-US) - Female".
 */

/**
 * @returns {Promise<Voice[]>}
 */
async getVoices() {
  // ...
}
```

src/providers/SpeechProvider/SpeechProvider.actions.js
sylvansson: Very minor detail but there are two spaces before the opening parenthesis instead of one.

src/components/Settings/Speech/Speech.component.js
```diff
@@ -77,7 +81,11 @@ const Speech = ({
               secondary={getVoiceLabel(voice)}
             />
           </ListItem>
-          <ListItem divider aria-label={intl.formatMessage(messages.pitch)}>
+          <ListItem
+            disabled={voice.voiceSource === 'cloud' ? true : false}
```
sylvansson, line 103: Maybe we could add a boolean property like `isCloud` to voice objects so that we can just do `disabled={voice.isCloud}`.

src/components/Board/Board.container.js
```diff
@@ -1458,18 +1459,28 @@ const mapStateToProps = ({
...
+  );
   const emptyVoiceAlert =
     speech.voices.length > 0 && speech.options.voiceURI !== EMPTY_VOICES
       ? false
       : true;
+  const offlineVoiceAlert =
+    !isConnected &&
+    speech.voices.length &&
+    currentVoice &&
+    currentVoice.voiceSource === 'cloud'
```
tomivm: I think that we can use `currentVoice?.voiceSource === 'cloud'` to reduce de `&&` concatenation

src/components/Settings/Speech/Speech.component.js
```diff
@@ -90,10 +98,14 @@ const Speech = ({
                 max={MAX_PITCH}
                 step={INCREMENT_PITCH}
                 onChange={handleChangePitch}
+                disabled={voice.voiceSource === 'cloud' ? true : false}
               />
             </div>
           </ListItem>
-          <ListItem aria-label={intl.formatMessage(messages.rate)}>
+          <ListItem
+            disabled={voice.voiceSource === 'cloud' ? true : false}
```
tomivm, line 124: if the user have selected an online voice and it is offline. `speech.voices` would not contain the selected voice and `voice` would be undefined. This throws an error. We can prevent this checking `voice?.voiceSource === 'cloud`.
the same in the others conditionals.
probably this isn't necessary if we prevent an undefined voice.

src/components/Settings/Speech/Speech.component.js
```diff
@@ -107,27 +119,35 @@ const Speech = ({
...
+          id="voice-menu"
+          anchorEl={anchorEl}
+          open={voiceOpen}
+          onClose={handleVoiceClose}
+        >
+          {langVoices.map((voice, index) => (
+            <MenuItem
+              key={index}
+              selected={index === selectedVoiceIndex}
+              onClick={() => onMenuItemClick(voice, index)}
```
tomivm, line 139: we should prevent the selection of an "online" voice if the redux state `app.isConnected` is false.
I mean if we don't prevent this. Voice would be undefined and the communicator hasn't voice
I think that is a good option show a notification and prevent the fire `onMenuItemClick(voice, index)`.
I could be wrong.

src/api/api.js
```diff
@@ -53,6 +55,24 @@ class API {
...
+    const azureVoicesListPath = `${AZURE_VOICES_BASE_PATH_API}list`;
+    const headers = {
+      'Ocp-Apim-Subscription-Key': AZURE_SPEECH_SUBSCR_KEY
+    };
+    try {
+      const { status, data } = await this.axiosInstance.get(
+        azureVoicesListPath,
+        { headers }
+      );
+      if (status === 200) return data;
```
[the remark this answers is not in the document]
martinbedouret: To me the check for 200 is necessary since we want to avoid returning any spurious data from Azure in case that a successful response different than 200.

src/providers/SpeechProvider/SpeechProvider.actions.js
martinbedouret: great feedback, will implement it in the next update

PR #965, opened 2021-07-01
Refactor redux state tour

### LINKED ISSUE #960: Refactor redux state for the new users tour
We want to Implement one status for any tour instead of a single status. So currently we have the following state variable in order to show / hide the tour: 
```
const initialState = {
...
  isTourEnabled: true,
...
```

And we want to make something like: 
```
const initialState = {
...
 liveHelp: {
    isRootBoardTourEnabled: true,
    isSettingsTourEnabled: true,
    isCommunicatorTourEnabled: true,
    isAnalyticsTourEnabled: true,
...
```
Final goal is to have a state for every tour to be implemented in the future.

RodriSanchez1:
State added for each tour inside cboard. 
All use one function `disableTour(payload)` to change the state depending which tour you saw


src/components/Board/Board.container.js
```diff
@@ -1463,7 +1463,12 @@ const mapStateToProps = ({
   communicator,
   speech,
   scanner,
-  app: { displaySettings, navigationSettings, userData, isTourEnabled },
+  app: {
+    displaySettings,
+    navigationSettings,
+    userData,
+    liveHelp: { isRootBoardTourEnabled }
```
martinbedouret: You should map the entire `liveHelp` object here

src/components/Board/Board.component.js
```diff
@@ -381,29 +381,29 @@ export class Board extends Component {
...
+              run={isRootBoardTourEnabled}
               styles={joyRideStyles}
               locale={{
                 last: <FormattedMessage {...messages.walkthroughEndTour} />,
                 skip: <FormattedMessage {...messages.walkthroughCloseTour} />
               }}
             />
           )}
-          {!isLocked && isTourEnabled && (
+          {!isLocked && isRootBoardTourEnabled && (
```
martinbedouret: since now we have a state for every tour, this element should use a different state than `isRootBoardTourEnabled`. You can use the name `isUnlockedTourEnabled`

martinbedouret (changes requested):

martinbedouret (approved):

PR #969, opened 2021-07-06
Fix of (Un)locking working not propperly during creating Boards

### LINKED ISSUE #711: (Un)locking is not working properly after trying to change board title on web Cboard
After trying to edit the board's title (#710 ), I noticed that toggle button for unlocking, when clicked, is showing the opposite information for how many clicks is needed for (un)locking.
The lock image is unlocked but it says there are more clicks needed to unlock the board (but actually it locks it and 1 click unlocks it).
(Lenovo laptop, Chrome browser)

image

tomivm:
this PR fix a double SWITCH_BOARD when the user is logged and creates a new empty board. this caused a wrong behavior on the 'clicks to unlock'
I'm not sure if is the same situation that the linked issue but was the only case that I could get the same described behavior.


martinbedouret (approved):
great job @tomivm !

PR #1014, opened 2021-08-17
Redirect to login page in case API returns 403 error 

### LINKED ISSUE #1015: Redirect to login page in case API returns 403 error

martinbedouret:


src/api/api.js
```diff
@@ -11,6 +12,8 @@ import {
 } from '../constants';
 import { getStore } from '../store';
 import { dataURLtoFile } from '../helpers';
+import { LOGOUT } from '../components/Account/Login/Login.constants';
```
tomivm: it's better to import the action directly? 
`import { logout } from '../components/Account/Login/Login.actions.js'`

sylvansson (approved):

src/api/api.js
martinbedouret: you are right, fixed

PR #1091, opened 2021-12-06
Feature/add tile dialog

### LINKED ISSUE #1061: New add tile dialog
The main goal of this feature is to simplify the addition of new tiles when the edit mode is enabled. By default, when a new board is created, Cboard will add a fluent board. Anyway, it is possible to transform the board into a fixed board. They have the following features:

The grid will not change for different screen sizes or device orientation.
The number of rows and columns are configurable.
The tiles will have fixed positions in the grid.
Hence, there are two scenarios for this feature:

1. **Fluent Board** 
addtile

addtile2

2. **Fixed Board** 
addtile3

addtile4

tomivm:
This feature allows users to add tiles Faster. 
If the board is fixed. is possible to add tiles on a wanted position.
A Keyboard plugin is needed for a proper merge of the feature. See the PR https://github.com/cboard-org/ccboard/pull/21


src/components/Board/TileEditor/TileEditor.component.js
```diff
@@ -488,7 +536,268 @@ export class TileEditor extends Component {
...
+        onChange={this.handleSymbolSearchChange}
+        onClose={this.handleSymbolSearchClose}
+      />
+    );
+
+    const parcialScreenContent = () => {
+      const { isKeyboardOpen, keyboardHeight } = this.state.keyboard;
+
+      const dialogContentStyle = () => {
+        if (isKeyboardOpen && isAndroid) {
```
tomivm, line 731: isAndroid its a function that should be called

tomivm:
This feature is awesome, but this branch has become too old to be maintained. It have so many conflicts now. It can be useful as a reference for a new PR to make #1061 
I prefer to close this and have a new PR to solve it

PR #1092, opened 2021-12-08
add style to live mode

### LINKED ISSUE #1051: text on Live feature is not properly aligned.
imagen
this cause more issues if the scroll bar is necessary
imagen

### LINKED ISSUE #1071: Live mode labels not visible in dark mode
Change to dark mode in the settings and try to add a symbol in live mode
Capture3
.

tomivm:
close #1071 close #1051


src/components/Board/Symbol/Symbol.css
```diff
@@ -7,6 +7,27 @@
   overflow: hidden;
   background: none;
 }
+.Symbol textarea {
+  align-self: flex-start;
+  text-align: center;
```
martinbedouret: I think that left align looks better

martinbedouret (changes requested):
awesome job, congrats @tomivm

tomivm:
Thanks for the review @martinbedouret . I made some changes to fix issues if 'Remove symbols from the output bar' is enabled. There is a test failing because I use [CSS vars](https://developer.mozilla.org/es/docs/Web/CSS/var()). I can't find a solution to disable that warning. It' looks like Csslint is outdated. you can see this [same issue](https://github.com/CSSLint/csslint/issues/720). Can you ignore the issue on Codacy? 
The browser support to CSS vars is really high. Only IE does not support it.

PR #1094, opened 2021-12-09
fixes for boards synch between browsers and between devices

### LINKED ISSUE #882: After I define a custom tile with my own picture in my iphone, the tile cannot be found in the browser in my laptop.
windows 10 , microsoft edge.

tomivm:
This PR is the init to solve the synch problems between devices
before this, the fn translateBoard() on componentDidMount of Board.container.js used an unsynched board like a parameter. close #882

### DIFF
--- a/src/components/Board/Board.container.js
+++ b/src/components/Board/Board.container.js
@@ -200,7 +200,6 @@ export class BoardContainer extends Component {
 
     const {
       board,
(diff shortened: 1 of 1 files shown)


martinbedouret (approved):

martinbedouret:
good!

PR #1097, opened 2021-12-10
Await to get voices before mount app. prevent double mounting

### LINKED ISSUE #1035: Double Mounting of App.component in every direct route access
if the app is accessed using the URL "https://app.cboard.io/" there is a double Mounting off the App.component. The same issue occurs in every browssing access to routes like "http://localhost:3000/login-signup"
imagen
imagen

tomivm:
close #1035

### DIFF
--- a/src/providers/SpeechProvider/SpeechProvider.container.js
+++ b/src/providers/SpeechProvider/SpeechProvider.container.js
@@ -23,35 +23,60 @@ export class SpeechProvider extends Component {
   state = { voicesRecived: false };
 
   async componentDidMount() {
-    const {
-      getVoices,
-      updateLangSpeechStatus,
-      getTtsEngines,
-      getTtsDefaultEngine,
-      ttsEngine,
-      setTtsEngine
-    } = this.props;
-    if (tts.isSupported()) {
-      //if android we have to set the tts engine first
-      if (isAndroid()) {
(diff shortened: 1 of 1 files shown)


src/providers/SpeechProvider/SpeechProvider.container.js
```diff
@@ -48,10 +50,12 @@ export class SpeechProvider extends Component {
       } catch (err) {
         console.error(err.message);
       }
+      this.setState({ voicesRecived: true });
     }
   }
 
   render() {
+    if (!this.state.voicesRecived) return null;
```
martinbedouret, line 83: @tomivm what would happen if the `await getVoices()` sentence does never return? Remember that js async calls don't have a default timeout, and getting voices depends on the platform. looks risky to render blank, it will Break the rendering chain of the application.

martinbedouret (changes requested):

src/providers/SpeechProvider/SpeechProvider.container.js
tomivm: thanks for your review! I will avoid it

tomivm:
Please @martinbedouret let me know the time limit that do you consider relevant to prepare voices.

martinbedouret (approved):

PR #1105, opened 2021-12-20
Manage switchs value on settings with enter key

### LINKED ISSUE #919: switchs on settings cannot change the value with the enter key
the switches on settings change value only with the 'space' key and not with 'enter' key. This is a bug for keyboard Users that only use 'tab' 'shift+tab' and 'enter'
image

tomivm:
close #919


tomivm:
Hey @shayc do you think this is useful?

shayc:
> Hey @shayc do you think this is useful?

Only if someone complained about this or you feel it is should support Enter key.
Personally I wouldn't touch it.

PR #1106, opened 2021-12-20
Fix/prevent login with a voice for another language

### LINKED ISSUE #1007: voiceUri value isn't sent to cboard-api when the user change Language
When logged user changes language. The default voice is not updated on user settings Data (database)

### LINKED ISSUE #1107: If voice of the user isn't stored on db. After login the app use a voice stored on redux without take care of user Language
The app should preserve the voiceUri stored on redux. Only if it was developed for the user language. If not should try to use the voiceURI stored on DB. If is unavailable should use the device's default voice.

tomivm:
close #1007


martinbedouret:
@tomivm I am not sure if we have to save the voice uri into the db. The voice uri depends on the device / browser you load the application. So if you save it for example using chrome and then you open the app in Firefox, the voice uri will not be there. The voice uri looks as a local parameter to me, we have to store it in the redux state, but not in the database.

tomivm:
@martinbedouret I Reopen the PR because I think in a solution after your comment. The only moment that the voice change without the user interacting is after login. Considering that already we store and get the voiceURI on db. I propose to check if app voice language (stored on the device) before login match with logged user language. if true, preserve it. if not Try to set the voice saved on db. If is unavailable set the device's default voice.
I think that could be a good idea to open a new issue. the already linked issue don't have a relation with this solution

src/components/Account/Login/Login.actions.js
```diff
@@ -22,6 +23,61 @@ export function logout() {
...
+    const browserVoiceLanguage = voiceLang?.substring(0, 2);
+
+    if (voices) {
+      const uris = voices.map(v => {
+        return v.voiceURI;
+      });
+      //if redux state have a defined voiceUri. Set it By default
+      if (
+        browserVoiceUri &&
+        browserVoiceLanguage === userLanguage &&
```
martinbedouret: This comparison is not a good idea. Most users are created in `en-US` and we never change that.

src/components/Account/Login/Login.actions.js
```diff
@@ -22,6 +23,61 @@ export function logout() {
...
+        const userVoiceLang = voices.filter(
+          voice => voice.voiceURI === userVoiceUri
+        )[0]?.voiceURI;
+
+        const userVoiceLanguage = userVoiceLang.substring(0, 2);
+
+        if (
+          userVoiceUri &&
+          userLanguage === userVoiceLanguage &&
+          uris.includes(userVoiceUri)
```
martinbedouret, line 64: same thing as before

martinbedouret (changes requested):

src/components/Account/Login/Login.actions.js
tomivm: The variable `userLanguage `  refers to the language.lang state. Should be renamed to `appLanguage`.

martinbedouret (approved):

PR #1140, opened 2022-02-10
Android - Share board on facebook

### LINKED ISSUE #769: Can't return to Cboard easily after publishing the board on Facebook
After the board is hared on Facebook, next screen shows:
Screenshot_20200803-145003_Cboard
It says to close this tab to continue but there is no any button to close this "tab"/screen/window. After clicking the back button on the device, we are back on the Facebook screen for sharing which also doesn't have a button to return to Cboard. It has an X but clicking this button return us to the previous screen.
Only if we click the back button on a device again, we will be returned to the Cboard (after a few seconds of a white screen).

RodriSanchez1:
This pull request will close: #769

It uses the recently added cordova facebook-connect plugin.
It allows you to share boards on facebook. It does not matter whether or not the user has the Facebook app installed on their device.

### DIFF
--- a/src/components/Board/BoardShare/BoardShare.component.js
+++ b/src/components/Board/BoardShare/BoardShare.component.js
@@ -36,27 +36,29 @@ function shareBoardOnFacebook(url, intl) {
     hashtag: '#Cboard',
     quote: intl.formatMessage(messages.subject)
   };
+
+  const errorFunction = msg => {
+    if (msg.errorCode != '4201')
+      alert(intl.formatMessage(messages.cannotShare));
+  };
+
   window.facebookConnectPlugin.logout(
     function succcesFunction(msg) {},
     function(msg) {
       console.log('error facebook disconnect msg' + msg);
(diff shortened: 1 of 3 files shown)


src/components/Board/BoardShare/BoardShare.component.js
```diff
@@ -27,6 +27,40 @@ import withMobileDialog from '@material-ui/core/withMobileDialog';
...
+function shareBoardOnFacebook(url, intl) {
+  const shareData = {
+    method: 'share',
+    href: url,
+    hashtag: '#Cboard',
+    quote: intl.formatMessage(messages.subject)
+  };
+
+  const errorFunction = msg => {
+    if (msg.errorCode != '4201')
```
martinbedouret: Expected '!==' and instead saw '!='.

martinbedouret (changes requested):

martinbedouret (approved):

PR #1147, opened 2022-02-23
Android - Support for universal link

### LINKED ISSUE #770: Facebook's browser for mobile devices can't open boards shared on Facebook
When clicked on the Facebook's shared board link, it automatically opens in the Facebook's browser for mobile devices and the page it is showing is blank, like in the screenshot:
Screenshot_20200803-150040_Facebook
After clicking on the settings and opening it in Samsung's browser, the board is shown.

RodriSanchez1:
Close #1140 

The new feature introduces the ability to open the Android app directly from a link which contains cboard host. If it is a board, the app will prompt the copy public board message o redirect to the existing board.
A new plugin was added for this feature [ionic-deep-linlk](https://github.com/ionic-team/ionic-plugin-deeplinks)

Also in index.html some meta tags were added to integrate this new feature with the Facebook app. If the user has installed the cboard app on their device, it will open . Otherwise, the Play Store will open with the cboard app listed.


src/cordova-util.js
```diff
@@ -7,7 +9,14 @@ export const isElectron = () =>
   isCordova() && window.cordova.platformId === 'electron';
 
 export const onCordovaReady = onReady =>
-  document.addEventListener('deviceready', onReady, false);
+  document.addEventListener(
+    'deviceready',
+    () => {
+      configDeepLinkPlugin();
```
martinbedouret: plugins should initialize in `initCordovaPlugins` function, not here.

src/components/Board/Board.container.js
```diff
@@ -278,14 +278,15 @@ export class BoardContainer extends Component {
     // if (isAndroid()) downloadImages();
   }
 
-  UNSAFE_componentWillReceiveProps(nextProps) {
+  async UNSAFE_componentWillReceiveProps(nextProps) {
```
martinbedouret: we should not grow this function, as it is going to be deprecated. See https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops
Place your code into another function.

martinbedouret (changes requested):

tomivm:
Is this still necessary? @RodriSanchez1

RodriSanchez1:
Yes! It's a good feature to open the Cboard app as default when you click on cboard board link. We should maintain the branch and  merge it soon

PR #1154, opened 2022-03-09
Fix service worker registration

RodriSanchez1:


martinbedouret:
@RodriSanchez1 check #1156

PR #1156, opened 2022-03-13
fix service worker, double mounting, and remove deprecated react lifecycle functions

### LINKED ISSUE #1153: WPA - Service Worker is not working
Service Worker is not installing in the web browser. Because of this, WPA is not available for download

martinbedouret:
This PR simplifies the logic for providers by fixing service worker, double mounting, and removing deprecated react lifecycle functions.

### DIFF
--- a/src/providers/SpeechProvider/SpeechProvider.container.js
+++ b/src/providers/SpeechProvider/SpeechProvider.container.js
@@ -14,7 +14,6 @@ import { isAndroid } from '../../cordova-util';
 
 export class SpeechProvider extends Component {
   static propTypes = {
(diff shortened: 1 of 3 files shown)


src/providers/SpeechProvider/SpeechProvider.container.js
```diff
@@ -23,74 +22,46 @@ export class SpeechProvider extends Component {
   state = { voicesRecived: false };
```
tomivm: this state is unnecessary

tomivm (approved):
Really good fix @martinbedouret

RodriSanchez1 (approved):

PR #1170, opened 2022-03-31
Full refactor for download language feature

martinbedouret:


src/components/Settings/Language/Language.container.js
```diff
@@ -403,13 +425,26 @@ export class LanguageContainer extends Component {
...
       ttsEngines,
       ttsEngine,
       history,
       showNotification
     } = this.props;
 
+    this.setState({ downloadLangLoading: false });
+
+    if (engineName === ttsEngine.name && localLangs.includes(selectedLang)) {
+      await setDownloadingLang({ isdownloading: false });
```
tomivm: setDownloadingLang is not an async function

src/components/Settings/Language/Language.container.js
```diff
@@ -87,10 +83,34 @@ export class LanguageContainer extends Component {
...
+    if (isdownloading) {
+      this.setState({ downloadLangLoading: true });
+    } else {
+      this.setState({ downloadLangLoading: false });
+    }
+  };
+
+  componentDidUpdate = async () => {
+    const { isdownloading, isUpdated } = this.props.downloadingLang;
+    if (isdownloading && isUpdated) {
```
tomivm, line 105: The condition should compare props with prevProps to prevent a infinite loop.
``` 
componentDidUpdate = async (prevProps) => {
    const { isdownloading, isUpdated } = this.props.downloadingLang;
    if(!isdownloading) return
    if (!prevProps.downloadingLang.isUpdated && isUpdated) { 
         await this.lookDownloadingLang();
    }
  };
```

src/providers/SpeechProvider/SpeechProvider.container.js
```diff
@@ -26,28 +30,38 @@ export class SpeechProvider extends Component {
...
       }
       if (ttsEngine && ttsEngine.name) {
+        const ttsEnginesName =
+          downloadingLang?.isdownloading &&
+          downloadingLang.engineName !== ttsEngine.name
+            ? downloadingLang.engineName
+            : ttsEngine.name;
         try {
-          await setTtsEngine(ttsEngine.name);
+          await setTtsEngine(ttsEnginesName);
```
tomivm, line 52: If the user has never selected a TTS Engine before. The corresponding tts engine for the selected language never is set.
we could remove this condition during download. Or preserve changes on setTtsEngines(). src/providers/SpeechProvider/SpeechProvider.actions.js:62 
to prevent that ttsEngine could be null.

src/components/Settings/Language/Language.container.js
```diff
@@ -341,6 +362,7 @@ export class LanguageContainer extends Component {
       engineName: ttsName,
       marketId: marketId,
       selectedLang: lang,
+      isUpdated: false,
```
tomivm, line 361: after this, the voices never would be updated because SpeechProvider.container never be remounted. Note that close the app and remount SpeechProvider.container is not necessary to change tts and Update the localLangs status with UpdateLangSpeechStatus()

src/components/Settings/Language/Language.component.js
```diff
@@ -205,7 +205,7 @@ class Language extends React.Component {
 
     const isDownloadable = avaliableAndDownloadablesLangs.filter(
       downloadableLang => {
-        return downloadableLang.lang === lang;
+        return downloadableLang.langCode === lang.slice(0, 2);
```
tomivm: this comparison cause dialects would be confuseds. 
Example: Sr-Rs and Sr-Me

src/providers/SpeechProvider/SpeechProvider.actions.js
```diff
@@ -52,26 +52,19 @@ export function receiveTtsEngine(ttsEngineName) {
...
-        .getTtsEngines()
-        .map(tts => tts.name)
-        .includes(selectedTtsEngineName);
-      const engineName = engineAvailable
-        ? selectedTtsEngineName
-        : tts.getTtsDefaultEngine().name;
-      const voices = await tts.setTtsEngine(engineName);
-      dispatch(receiveTtsEngine(engineName));
+      const voices = await tts.setTtsEngine(ttsEngineName);
+      dispatch(receiveTtsEngine(ttsEngineName));
```
tomivm, line 67: a deleted or unavailable tts could be set

src/components/Settings/Language/Language.container.js
```diff
@@ -87,10 +83,34 @@ export class LanguageContainer extends Component {
...
+        : {
+            avaliableAndDownloadablesLangs: [],
+            downloadablesOnly: []
+          }
+    });
+    if (isdownloading) {
+      this.setState({ downloadLangLoading: true });
+    } else {
+      this.setState({ downloadLangLoading: false });
+    }
```
tomivm: this could be replaced with 
`this.setState({ downloadLangLoading: isdownloading });`

src/components/Settings/Language/Language.container.js
```diff
@@ -418,86 +453,16 @@ export class LanguageContainer extends Component {
...
-          downloadingLangError: {
-            ttsError: false,
-            langError: true
-          }
-        });
-      }
-    }
-    const localLangs = this.props.localLangs;
-    if (!localLangs.includes(selectedLang)) {
+    } else if (
```
tomivm, line 452: return if (!ttsEnginesNames.includes(engineName)) 
and remove else if condition. if any of before conditions would be true only set:
```
        downloadingLangError: {
          ttsError: false,
          langError: true
        }
```

tomivm (changes requested):

src/components/Settings/Language/Language.container.js
martinbedouret: fixed

src/providers/SpeechProvider/SpeechProvider.container.js
martinbedouret: that could never happen as there is a condition on line 45:
```
     if (ttsEngine && ttsEngine.name) {
```

src/components/Settings/Language/Language.component.js
martinbedouret: fixed

src/components/Settings/Language/Language.container.js
martinbedouret: done

src/components/Settings/Language/Language.container.js
martinbedouret: can you rephrase your comment? I don't understand.

src/components/Settings/Language/Language.container.js
martinbedouret: this is intentional in order to allow asynchronous behavior between `speechProvider` and this component. An infinite loop should not happen as we are removing the prop once it was used.

src/components/Settings/Language/Language.container.js
martinbedouret: I don't see the point here. When user changes the tts engine manually from the language settings, there is no need to update the speech container .

src/providers/SpeechProvider/SpeechProvider.actions.js
martinbedouret: yes, but that should be enforced by the upper logic and not by the action. Ideally, an action must abstract the reducer access and doesn't apply business / logic  validation.

martinbedouret:
@tomivm Great review!! Thanks

src/components/Settings/Language/Language.container.js
tomivm: Sorry. If ttsEnginesNames does not include the engine name we could return from the function after set ttsError to true. if not, We know that is a langError and the else if condition would be ambiguous.

src/components/Settings/Language/Language.container.js
tomivm: I mean that the function is to change tts engine and set the voice through the 'configure a local voice' button. the point is that now. the tts engine never changes and the app is stoped on the loading view.

src/providers/SpeechProvider/SpeechProvider.container.js
tomivm: Sorry, I will try to explain it better. The point is that if ttsEngine is null (is the case that the user never selects a tts manually with the tts selector before download) The engine for the downloading lang is never set. I mean all the bunch of code is wrapped on `     if (ttsEngine && ttsEngine.name) { ` 
In resume. A null TTS engine before downloading is causing that `downloadingLang.engineName` didn't set automatically

src/components/Settings/Language/Language.container.js
tomivm: Im not sure if we are removing the prop before a componentDidUpdate is fired again. componentDidUpdate() is fired after every change. Like a setState(). one of the first lines on lookDownloadingLang() is `this.setState({ downloadLangLoading: false });`  and this for example is firing componentDidUpdate again. 
see src/components/Settings/Language/Language.container.js:435 and https://es.reactjs.org/docs/react-component.html#componentdidupdate
now we are getting this error: `Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.`
I think that the change recommended is not breaking the asynchronous behavior between the Speech Provider and this component because lookDownloadingLang() is fired inmediatly after voices and langs are updated.

tomivm (commented):
Thank you for showing me how to improve my code! Eternally grateful :)

src/providers/SpeechProvider/SpeechProvider.actions.js
tomivm: thanks for your explanation.

src/components/Settings/Language/Language.container.js
tomivm: Like Mostly times. You are on truth. On the new PR #1229 I refactor the function to set isUpdated: false immediately after using it and the loop is not executed.

PR #1173, opened 2022-04-05
Feature/increase output UI size

### LINKED ISSUE #974: Increase UI component size for accesibility
Based on user feedback and from some tests with a new accessibility device, we would like to increase the size and/or "clickable area" for specific components in the UI. 

We are testing an eye tracker and the precision is not enough to use the current UI size. 

**What have we tested?**
We already tried increasing the size of the UI component from:
Display -> Increase UI size
Display -> Increase Font size

We suggest that this size should be at least 66% of the size of a tile. (Only for this accessibility device, but we suggest to increase it for all users, especially the ones with motor impairments).

Buttons 1 and 2 are crucial for usability. 
Captura de Pantalla 2021-07-13 a la(s) 14 52 22

tomivm:
related to #974


src/components/Board/Output/SymbolOutput/SymbolOutput.css
```diff
@@ -54,6 +54,73 @@
...
+  height: 100%;
+  justify-content: center;
+  align-items: center;
+}
+
 .SymbolOutput__right__btns .MuiFormControlLabel-label {
   font-size: 11px;
 }
+
+.Output__button__sm {
```
RodriSanchez1: add the button tag before .Output_button__sm like the others classes

src/components/Board/Output/SymbolOutput/ClearButton/ClearButton.js
```diff
@@ -27,16 +15,26 @@ export class ClearButton extends Component {
...
+            }
+          />
         </IconButton>
       </Scannable>
     );
   }
 }
 
-export default withStyles(styles)(ClearButton);
+export default withStyles(null)(ClearButton);
```
RodriSanchez1, line 40: You can export directly the componente without the withStyles, like `export default (ClearButton)` and also deleate the import.

RodriSanchez1: You can export directly the componente without the withStyles, like `export default (ClearButton)` and also deleate the import.

RodriSanchez1 (changes requested):

src/components/Board/Output/SymbolOutput/ClearButton/ClearButton.js
```diff
@@ -27,16 +15,26 @@ export class ClearButton extends Component {
   };
 
   render() {
-    const { classes, hidden, ...other } = this.props;
+    const { classes, hidden, increaseOutputButtons, ...other } = this.props;
```
RodriSanchez1: You can delete the prop classes

src/components/Board/Output/SymbolOutput/SymbolOutput.css
tomivm: why?

martinbedouret (changes requested):
` Warning: A component is changing an uncontrolled input to be controlled. This is likely caused by the value changing from undefined to a defined value, which should not happen. Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://reactjs.org/link/controlled-components
`

martinbedouret (approved):

PR #1176, opened 2022-04-11
Correct differentiation between local and cloud voices

### LINKED ISSUE #1052: Browser online Voices are not detected like 'online'
This Causes that the user doesn't know that the voice is online and that the 'online voice Warning' does not appear if there isn't a connection.
imagen

RodriSanchez1:
In this pr the classification of online and local voices has been corrected.
We get the localService property to differentiate if the voice is installed on the device or not.
As issue #1052 indicates, we were treating the voices provided by the browser as local, when in fact they were online.
For android, the cboard-speech-tts plugin has been modified to obtain the feature property that tells us if the voice provided by android.tts.speech is installed on the device. In order to test this update in cordova, first you have to delete the cboard-speech-tts plugin and modify the package.json and package-lock.json files by changing the path of said plugin to `"cboard-speech-tts": "github:RodriSanchez1/cboard-speech-tts#fix/localVoices "` on ccboard repository.

close #1052


src/providers/SpeechProvider/SpeechProvider.actions.js
```diff
@@ -177,19 +178,35 @@ export function getVoices() {
...
             voice.lang = lang;
           } else if (Locale) {
             voice.lang = Locale;
           }
-          if (voiceURI) {
+          if (isAndroid() && features && /notInstalled/.test(features)) {
+            voice.voiceURI = voiceURI;
+            voice.voiceSource = 'cloud';
+            voice.isAzure = false;
+          } else if (voiceURI && localService) {
```
martinbedouret, line 210: @RodriSanchez1 The meaning of the `localService` properly **is not that a voice is from a cloud service or not.** The meaning of the property indicates if the voice comes from a remote service or not. In this context, a voice provided by the browser is remote, while a voice provided by the operating system will be a local voice. 
https://wicg.github.io/speech-api/#dom-speechsynthesisvoice-localservice

image

martinbedouret (changes requested):

src/providers/SpeechProvider/SpeechProvider.actions.js
RodriSanchez1: https://developer.chrome.com/blog/web-apps-that-talk-introduction-to-the-speech-synthesis-api/

martinbedouret (approved):

PR #1184, opened 2022-04-28
Feature/include adsense on settings

tomivm:
To test on localhost with an authorized connection to Adsense resources:  https://stackoverflow.com/a/64694852/16816650


martinbedouret (approved):

RodriSanchez1:
resolve #1204

PR #1214, opened 2022-05-18
feature - navigate on fixed boards with arrow keys

### LINKED ISSUE #1095: Add the funcionabillity to navigate between tiles whith the arrows keys

tomivm:
close #1095

### DIFF
--- a/src/components/FixedGrid/Grid.js
+++ b/src/components/FixedGrid/Grid.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 
@@ -16,6 +16,10 @@ function chunks(array, size) {
   return results;
 }
 
+const focusPosition = {
+  x: 0,
+  y: 0
+};
 function Grid(props) {
   const { className, items, style, ...other } = props;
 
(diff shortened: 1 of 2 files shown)


RodriSanchez1:
-  The focus is lost if open a  fixed board  and that dont have a tile on that position.
- If I press an arrow while a fiexed board is open, a tile should be in focus

src/components/FixedGrid/Grid.js
```diff
@@ -24,8 +28,121 @@ function Grid(props) {
...
+
+    return () => {
+      window.removeEventListener('keyup', tileHasFocus);
+      window.removeEventListener('keydown', preventScroll);
+    };
+  }, []);
+
   return (
-    <div className={styles.root} style={style}>
+    <div className={styles.root} style={style} onKeyDown={handleOnKeyUp}>
```
RodriSanchez1: The name of the event is `onKeyDown`. why you call the function `handleOnKeyUp`? They  are differents events . See [onKeyDown vs onKeyUp](https://javascript.info/keyboard-events)

src/components/FixedGrid/Grid.js
```diff
@@ -24,8 +28,121 @@ function Grid(props) {
...
+
+    const compareKeys = code => {
+      const right = code === 'ArrowRight';
+      const left = code === 'ArrowLeft';
+      const up = code === 'ArrowUp';
+      const down = code === 'ArrowDown';
+
+      if (!(right || left || up || down)) return true;
+      event.preventDefault();
+
```
RodriSanchez1: If you move the conditional `if (event.repeat) return;`  used on line 34 to the line 44, dont need to add the listener for prevent scroll on line 136

src/components/FixedGrid/Grid.js
```diff
@@ -24,8 +28,121 @@ function Grid(props) {
...
+        if (!firstTile) return;
+        firstTile.focus();
+        const firstTilePosition = firstTile.parentNode.parentNode.id;
+        focusPosition.x = parseInt(firstTilePosition[0]);
+        focusPosition.y = parseInt(firstTilePosition[2]);
+      };
+
+      if (!isArrowKey(event)) return;
+
+      if (event.repeat) return;
```
RodriSanchez1: This is not necessary because the listener is on the onKeyUp event which occurs when the user releases a key

src/components/FixedGrid/Grid.js
```diff
@@ -24,8 +28,121 @@ function Grid(props) {
...
+  useEffect(() => {
+    const isArrowKey = event => {
+      const code = event.code;
+      const right = code === 'ArrowRight';
+      const left = code === 'ArrowLeft';
+      const up = code === 'ArrowUp';
+      const down = code === 'ArrowDown';
+
+      if (!(right || left || up || down)) return false;
+      event.preventDefault();
```
RodriSanchez1: These lines of code are the same used on lines 37 - 43, i suggest to refactor the functions!

RodriSanchez1 (changes requested):

src/components/FixedGrid/Grid.js
tomivm: good eye.. I start writing the code with onKeyUp event but later i start to using onKeyDown to can prevent default.

martinbedouret:
Some bugs: 

- Navigation doesn't work in edit mode for the board 
- Navigation doesn't scroll down if it is a large board

martinbedouret (changes requested):
Some bugs:

Navigation doesn't work in edit mode for the board
Navigation doesn't scroll down if it is a large boa

src/components/FixedGrid/Grid.js
tomivm: Appear a good change to only prevent event repeat for arrow keys. 
the only against removing the preventDefault of onKeyDown event handler is when the user clicks on the navbar. By pressing arrow keys scrolling occurs. 
If we want to save this behavior, we should prevent set focus on the first tile when the focus is on the Board body. 
if not we should prevent scroll

src/components/FixedGrid/Grid.js
tomivm: Cert! I will refactor the function and use onKeyDown to prevent scroll on the same event handler

src/components/FixedGrid/Grid.js
tomivm: Already refactorized.

tomivm:
@martinbedouret Thanks for your review :) Edit mode is working now.
 Navigating on larger pages too.
 About that, I'm not sure if fixed boards should be scrollable. I mean If users can scroll. The position of tiles related to the screen could be different and the board is not finally "fixed".

PR #1229, opened 2022-06-10
Fix/full refactor downl lang review

### LINKED ISSUE #1166: Improve 'download language' feature
- Offer the functionality to cancel the downloading process during the loading view.
- After pressing the 'configure a local voice' button should select the correct language if it is available. (it is working but sometimes the selected language state is losted and the user needs to click the language again).

tomivm:
I reviewed and refactor starting from https://github.com/cboard-org/cboard/pull/1170 
close #1166


src/providers/SpeechProvider/SpeechProvider.container.js
```diff
@@ -25,29 +29,65 @@ export class SpeechProvider extends Component {
...
       }
-      if (ttsEngine && ttsEngine.name) {
+
+      const ttsEngineName = getTtsEngineName();
+
+      if (ttsEngineName) {
         try {
-          await setTtsEngine(ttsEngine.name);
+          await setTtsEngine(ttsEngineName);
+          forceChangeVoice = true;
```
RodriSanchez1: When the apps starts, the voices always will be updated with the default voice of the tts. The variable `forceChangeVoice` needs to be true only when a new tts was dowloaded.

tomivm: truth! thanks

tomivm:
Taking advantage of the refactor instance. I implement an event handler for the Android back button. that before was ignored causing the possibility to navigate without canceling the downloading process.

tomivm:
Please, @gonojuarez can you fully test this? It is an Android-only feature. Let me know if you have doubts. Also, test the web app to make sure that all continues working correctly :)

gonojuarez:
I tested it, and it works well in Android 13.
I find problems with CBOARD TTS related to the following languages:
- Croatian
- Servian
- Montenegro
Those are available for Android devices running version 12 or earlier.

To run a TTS Voice Engine, you need to open the app twice.
First, searched the TTS engine and then enabled it as a voice.
Those voices are available in both modes, "Online " or "Offline"

Issue #1067, opened 2021-11-13
Missing symbols in offline mode


mzilak:
1) When starting Cboard without internet, boards created before show some symbols missing.
2) When internet is off during Cboard usage, first shown/used board is loaded normally, but choosing another board also doesn't load all the symbols.
3) Add new symbol from Mulberry, Global Symbols and ARASAAC gallery, go offline and open Cboard again - ARASAAC symbol will be missing (not loaded) while Mulberry and GD symbols are loaded normally.

Cboard (v. 1.15.3) used with Croatian language and AlfanumTTS Lite CRO.

![Screenshot_20210602-174339_Cboard](https://user-images.githubusercontent.com/57595384/141612088-2a814149-3f2b-4ebc-ade9-d7191801e1f4.jpg)
![Screenshot_20210602-174339_Cboard2](https://user-images.githubusercontent.com/57595384/141612091-66b8da42-0717-4644-b9f9-7e1d2e670174.jpg)

.

tomivm:
Thanks, @mzilak for the report!

I note that only symbols under the public/symbols folder are available on offline devices.

RodriSanchez1:
Duplicate of #795

eadraffan:
So sorry but I consider this a bit of a bug or issue rather than a feature to be voted on? I seem to remember this was mentioned back in the days our testing at the beginning.  I think it is important but I really want to put higher percentages by some other features!  ;>))

Issue #795, opened 2020-10-07
When offline not all images are shown


timnis:
I have make own board and its working fine when online. 

But when you offline it only shows some images but speak correctly.
Same problem on iPad/Safari, Android/Firefox, Android/Cboard apps. All my own images are not shown and major part of symbols not shown also

Here is screenshot from my OnePlus 7 Pro with Cboard apps, online and offline
![image](https://user-images.githubusercontent.com/31338854/95321310-90fb1e80-08a3-11eb-9b1c-d6be66fcf8d8.png) ![image](https://user-images.githubusercontent.com/31338854/95321266-79239a80-08a3-11eb-89ab-7fb6e794e71e.png)

martinbedouret:
Thanks for the work here @timnis !
Yes, we are aware of this issue and hope to resolve it soon. It will require some work.

timnis:
@martinbedouret  thanks for information.

PR #1237, opened 2022-07-07
Persistence of Arasaac symbols when device is offline

tomivm:


src/components/Board/SymbolSearch/SymbolSearch.component.js
```diff
@@ -232,14 +232,21 @@ export class SymbolSearch extends PureComponent {
 
   handleSuggestionsClearRequested = () => {};
 
-  handleSuggestionSelected = (event, { suggestion }) => {
+  handleSuggestionSelected = async (event, { suggestion }) => {
```
martinbedouret, line 235: I don't think this is the right place to make the update, as you will be updating all the suggestions (including global symbols). You should make the update on the fetch arasaac suggestions level.

martinbedouret (changes requested):

martinbedouret (approved):

RodriSanchez1:
close #1067
close #795

PR #1249, opened 2022-07-21
Localize user

### LINKED ISSUE #1247: Obtain the geolocation of users
The goal is to get the geolocation of users and send it to cboard API.  In the case of existing users, we have to detect when a user has an active session and then, get the location. 
Location is limited to the user's country. 

### RESOURCES
based on these Articles:
https://www.softwaretestinghelp.com/best-ip-geolocation-api/
https://stackoverflow.com/a/35123097
Consider the uses of the following resources:

1. https://www.geoplugin.com
prices to get SSL: https://www.geoplugin.com/webservices/ssl

:+1: 

- Free Acces without a token.
- High precision of city.

:-1: 

- No SSL connection without a paid account

2. https://geolocation-db.com

:+1: 

- Free Acces without a token.
- SSL connection without a paid account

:-1: 

- Low precision of city.

3. https://ipdata.co/
prices: https://ipdata.co/pricing.html

:+1: 

- 1500 free requests daily with a token. (should be stored on API)
- SSL connection with a free account
- High precision of city.

:-1: 

- is necessary a token that should be stored on our API.

5. https://ip-api.com/
Commercial use prices https://members.ip-api.com/#pricing

:+1: 

- 1500 free requests daily with a token. (should be stored on API)
- SSL connection with a free account
- High precision of city.

:-1: 
- Non-commercial use for free.
- No SSL connection without a paid account.

### When should store data?
-Should store the location when a Cboard user creates a new account.
-In the case that users that already have accounts should evaluate when make the requests.
       Consider checking if the location is available at a login moment. Compare if it is available to make the request or not

RodriSanchez1:


src/components/App/App.actions.js
```diff
@@ -46,3 +48,55 @@ export function updateUserData(userData) {
...
+          id: id,
+          location: {}
+        });
+        return userLocation;
+      };
+
+      if (location) return;
+      const userLocation = await APIGetAndUpdateLocation();
+      if (userLocation)
+        return dispatch(
```
martinbedouret: This looks strange, a function should not return a dispatcher. Although it works, it would be more readable if first dispatch and then return.

martinbedouret (commented):
great job!! Just a little comment.

tomivm:
Thanks for your review!

PR #1264, opened 2022-09-23
Feature/fix touch bug

### LINKED ISSUE #1233: Tile is not pressed if user hold tap
holding a tile pressed results in selecting the tile and not submitting a word. This behavior could be wrong for users that have motor impairments.

vlaco:
The bug with multi touch was fixed by installing hammerjs library. Please test if everything is still working properly. An attempt to resolve https://github.com/cboard-org/cboard/issues/1233


martinbedouret:
@vlaco Thanks for this. Is there an open issue with the bug?
Please remove lock files from the PR.
@tomivm Could you please do the review for this?

tomivm:
thanks @vlaco ! please link the issue that you are resolving.
About the testing. 
Tiles aren't rendered and I receive the following errors on the console:
```
Warning: Each child in a list should have a unique "key" prop.

Check the render method of `Board`. See https://reactjs.org/link/warning-keys for more information.
```
and

`Warning: Failed prop type: Duplicate child key "null" found! This will cause problems in ReactGridLayout.`

src/components/Board/Board.component.js
```diff
@@ -192,33 +192,29 @@ export class Board extends Component {
       const variant = Boolean(tile.loadBoard) ? 'folder' : 'button';
 
       return (
-        <div key={tile.id}>
```
tomivm: The key is necessary. See https://reactjs.org/docs/lists-and-keys.html

tomivm (changes requested):

vlaco:
> @vlaco Thanks for this. Is there an open issue with the bug? Please remove lock files from the PR. @tomivm Could you please do the review for this?

@martinbedouret this is an attempt to resolve this issue https://github.com/cboard-org/cboard/issues/1233

RodriSanchez1:
Hi! @vlaco and @matijavibe! Thanks for been part of these amazing project!
I just tested in web and android device. When you try to scroll the board view in touch devices  the tapped tile is pressed.  You will need to treat the pan event.

tomivm:
I note the same as @RodriSanchez1 . When the users scroll the board tiles are pressed. Also, this PR eliminates the possibility of selecting tiles with the scanner feature and the feature of navigating fixed boards with the arrow keys is damaged too.

vlaco:
In order to solve this a refactor of the whole app is needed

PR #1290, opened 2022-11-07
Decoupling constants and services

lutangar:
> see https://github.com/cboard-org/cboard-api/pull/233

### DIFF
--- a/dev.Dockerfile
+++ b/dev.Dockerfile
@@ -1,0 +1,16 @@
+# development
+FROM node:16-alpine AS development
+ENV NODE_ENV development
+# Add a work directory
+WORKDIR /app
+# Cache and Install dependencies
+COPY package.json .
+COPY yarn.lock .
+RUN yarn install
+# Copy app files
+COPY .. .
+
+# Expose port
+EXPOSE 3000
+# Start the app
+CMD [ "yarn", "start" ]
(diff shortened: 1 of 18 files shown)


martinbedouret:
@lutangar  thanks a lot for the PR. After talking to the team about this work we decided to don't merge this decoupling to the repository. The main reason is the time constraint we have to review and test this, and the associated risk to introduce new bugs. We are ok if you apply these changes in your forked repository. Looping @RodriSanchez1 @tomivm

lutangar:
Hello @martinbedouret and thank you for your feedback !

This PR wasn't meant to be merge as is anyway... but now I wonder, maybe have you missed the comment I made on the [related PR on the API](https://github.com/cboard-org/cboard-api/pull/233) ?

> [...] 
> I'm willing to propose each changes separately if you're interested in some or all of these changes.
> Each commit represent a feature. [...]
> https://github.com/cboard-org/cboard-api/pull/233

The same apply to these commits / features. Even if only some of them reach the __CBoard__ repository, I believe it would already be a small victory toward __CBoard__ "self-hostability" and thus a step toward decentralization.

That's why I'm taking a second chance and respectfully ask again : which features would you be willing to consider if submitted as a stand-alone ? *And of course, I'll be the one responsible to submit the testing and documentation part as well.*

cc @RodriSanchez1

PR #1300, opened 2022-11-28
Action to replace default communicator

tomivm:
It uses a default board only as an example.


tomivm:
This must be closed because a change of communicator is not required to change the default board feature.

PR #1302, opened 2022-12-02
Implemented option to export PDF board with PicseePal dimensions

### LINKED ISSUE #1126: Export fixed board with PicseePal dimension
Picseepal is a  low-tech visual communication system. This is a new feture which brings the possibilitty to export a fixed board with the Picseepal dimension.

Size of the page: Letter (81/2 × 11 pulgadas) | 279,4 mm × 216 mm
Dimension of the PicseeSnap : 199.39 mm x 132.588 mm
For best result , leave a 1/4 inch (6.35 mm) border of empty space between the edge of the board image and the inner most border of the PicseeSnap
For more information; https://docs.google.com/presentation/d/16MStzDuSiAChZ2Yjn5qPLowpXGxbg99dNPbywhcX1v0/edit#slide=id.ge3ef8c81d3_0_8

image

Expected result:

WhatsApp Image 2021-08-26 at 9 39 45 AM

dh-verma:
As requested in #1126, I’ve implemented PDF generation with PicseePal dimensions and have tested it with multiple boards. The PicseePal PDF is generated on standard A4 sized paper. The icons are scaled down and the PDF has guides to cut the paper to the right size to fit within the PicseePal device (200mm x 132mm viewable area). The solid line on the PDF represents the viewable area and the dashed line is where the paper should be cut.

Provided the user sets the scale in the print dialog to 100%, the dimensions will also be exactly correct when printed on US Letter sized paper (standard paper size in the US). Setting the scale to 100%, and NOT fit/shrink to fit, will ensure the print will be the correct dimensions regardless of the printer paper size.

If the user forgets to set the scale to 100%, then the size will still be exactly correct on A4 paper, but will be slightly smaller (~92%) on US letter paper. This reduction is not large, so it can still be used in a PicseePal (with a few mm extra white border).

Please let me know if you would like this implemented any differently. I also considered adding an option to ask the user if they would like the pdf in A4 or US Letter size, but I thought this may add confusion.

### DIFF
--- a/src/components/Settings/Export/Export.helpers.js
+++ b/src/components/Settings/Export/Export.helpers.js
@@ -508,23 +508,51 @@ const addTileToGrid = async (
     alignment: 'center',
     width: '100'
   };
-  if (picsee) {
-    // scale down images to fit inside PicseePal dimensions
-    imageData.width = '60';
-  }
 
   const labelData = {
     text: label,
     alignment: 'center'
   };
 
-  if (11 === columns || columns === 12 || rows >= 6) {
-    imageData.width = '59';
-    labelData.fontSize = 9;
-  } else if (9 === columns || columns === 10 || rows === 5) {
-    imageData.width = '70';
-  } else if (7 === columns || columns === 8) {
-    imageData.width = '90';
+  if (picsee) {
+    // This scales down images to fit inside PicseePal
+    // dimensions depending on number of columns
+    var colImgWidths = {
+      1: 130,
+      2: 130,
+      3: 80,
+      4: 80,
+      5: 75,
+      6: 60,
+      7: 55,
+      8: 55,
+      9: 45,
+      10: 45,
+      11: 40,
+      12: 37
+      // max num of columns or rows is 12
+    };
+    var colFontSizes = {
+      9: 9,
+      10: 9,
+      11: 8,
+      12: 7
+    };
+    imageData.width = colImgWidths[columns];
+    if (columns in colFontSizes) {
+      labelData.fontSize = colFontSizes[columns];
+    }
+    console.log(columns);
+  } else {
+    // if not picseepal PDF, then retain old method for computing image widths
+    if (11 === columns || columns === 12 || rows >= 6) {
+      imageData.width = '59';
+      labelData.fontSize = 9;
+    } else if (9 === columns || columns === 10 || rows === 5) {
+      imageData.width = '70';
+    } else if (7 === columns || columns === 8) {
+      imageData.width = '90';
+    }
   }
 
   const displaySettings = getDisplaySettings();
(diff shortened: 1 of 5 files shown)


dh-verma:
Regarding testing:
I tried to run Export.test.js, but it is failing as I have edited the UI (added a new option in export dropdown) so the new UI no longer matches the stored snapshot. Would you be able to advise how I can get this test to pass? I am not too familiar with snapshots, does the stored snapshot need to be updated with the new UI?

src/components/Settings/Export/Export.helpers.js
```diff
@@ -491,6 +508,10 @@ const addTileToGrid = async (
     alignment: 'center',
     width: '100'
   };
+  if (picsee) {
+    // scale down images to fit inside PicseePal dimensions
+    imageData.width = '60';
+  }
```
RodriSanchez1: You will need to add diferents `imageData.width` depending on the rows and columns of the board. This is made in line [500](https://github.com/cboard-org/cboard/pull/1302/files#diff-a5187bfb1603e0959ee1c612e12529bf7f9b80d2b8a56d5e8f7a8b390e9d058bR521) .  You can export a fixed board of 5 columns and 5 rows and will see  the last row is in other page. If the `imageData.width = '47' ` is exported correctly.

src/components/Settings/Export/Export.helpers.js
```diff
@@ -781,6 +802,97 @@ export async function pdfExportAdapter(boards = [], intl) {
   }
 }
 
+export async function picseePdfExportAdapter(boards = [], intl) {
```
RodriSanchez1: Can you refactor the `pdfExportAdapter` function to avoid innecesary code repetition? Note that the function is also used in `PrintBoardButton.container.js`

RodriSanchez1 (changes requested):
Good work!! @dh-verma  I left you some comments . My discord user is Rodri Sanchez#0410  . If you need some help or explain about the review or want to know more about the project send me an DM

To update the snapshot you can check [link](https://jestjs.io/docs/snapshot-testing#interactive-snapshot-mode)

src/components/Settings/Export/Export.helpers.js
dh-verma: I have refactored `pdfExportAdapter` in my newest commit to remove unnecessary code repetition.

src/components/Settings/Export/Export.helpers.js
dh-verma: I have implemented dynamic resizing of images based on the number of columns, in my newest commit. I have also tested PicseePal PDF exporting with all column sizes (1 through 12) to ensure nothing is cut off.

dh-verma:
Hi @RodriSanchez1 , I've implemented the feedback you provided.

List of changes in new commits:
1. refactored `pdfExportAdapter` and removed `picseePdfExportAdapter`
2. Added dynamic resizing for symbol images and font size based on number of columns in board
3. updated snapshot for Export.tests.js so test is now passing with new changes

I also tested exporting fixed boards with all possible column sizes (1-12) into PicseePal PDF to ensure nothing was cut off.Screenshot 2022-12-02 at 4 49 03 PM

RodriSanchez1:
Hello @dh-verma! We are almost there! I tried the feature and still have the problem that the board is cutting off. You are taking care of the columns but not the rows. For example, if I export a fixed board that has 5 rows and 5 columns, it gets cut off.

image29

dh-verma:
Hi @RodriSanchez1 , thanks for your feedback. I've implemented image resizing based on both rows and columns. Output now matches with your expected image.

martinbedouret:
Amazing work @dh-verma ! Congratulations! This will be published in the next release this week.

PR #1303, opened 2022-12-06
Adding Drag and Drop Functionality for Symbol Tiles in Output Bar (Issue #890)

### LINKED ISSUE #890: Feature: reordering elements in the output box
I thought about rearranging items in sentence (using drag & drop) in case of misclick or user fault.

dmaraffi:
In this fork we added drag and drop functionality for symbol tiles loaded into the output file as a response to the open issue feature request with id #890 . When the user holds the mouse down on one of the tiles in the output bar, the tile becomes slightly transparent and begins to move with the user’s mouse while they continue to hold the mouse down. If the user drags the tile and drops it past the middle point of the width of another tile, it will adjust the order to put the moved tile in a new spot, shifting the other tiles in one direction to accommodate the new order. Users are able to drag and drop tiles in both directions to be able to easily correct any mistakes made during their initial selection, rather than needing to delete the entire incorrect portion and rebuild it. In order to increase our confidence that our alterations worked as expected, we found individuals who had not worked on the code with us and asked them to try out the feature without any knowledge of how it worked internally to simulate how actual users would approach it and try to find any edge cases we may have missed.


martinbedouret:
@tomivm  can you please take a look at this?

PR #1312, opened 2022-12-12
added a loader during the initial app start-up, issue #1108

### LINKED ISSUE #1108: Show a Loading Screen (UI) During initiation proceses
The time for the initiation process could be different depending on devices, TTS, internet connection, and so on. We could develop a screen to show during this process.

l3atjin:
This change aims to add a loading animation while the app starts up. Any feedback is greatly appreciated!

### DIFF
--- a/public/index.css
+++ b/public/index.css
@@ -1,0 +1,33 @@
+.loader {
+  border: 16px solid #f3f3f3;
+  border-top: 16px solid #3498db;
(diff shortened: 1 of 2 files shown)


martinbedouret:
@RodriSanchez1  could you write some comments here? For me, the implementation should be at react components level and not in the html.

Issue #1323, opened 2022-12-19
ARASAAC offline support


shayc:
## Feature
Download and use ARASAAC symbol-set locally (on device - works offline)

### Technical description
The symbols will be provided via `cboard-api`, the entire zipped file weighs ~280MB, we should probably split it into smaller parts since unzipping in the browser takes between a couple of seconds to minutes depending on the file size and device.

Unzipping should be done async to avoid blocking the UI, need to verify if [`jszip`](https://stuk.github.io/jszip/) and/or `create-react-app` supports [`Web Workers API`](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API).

Storage should be done with [`IndexedDB API`](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API).
A new mechanism for displaying images from IndexedDB (board / image search).
Since searching for images will be done on the client, we need an index (json) of all ARASAAC images and their names in all the languages we support.

### Notes
* ~~IndexedDB size limit on different browsers?~~ seems like 1GB per origin at the very least [ref](https://web.dev/storage-for-the-web/)
* [idb](https://www.npmjs.com/package/idb) is a nice package for working with IndexedDB.
* does `jszip` has WebWorker support?
* how do we manage interrupted downloads (to avoid re-downloading the same package when resuming)
* need an index of all symbols and their names (also localization if we want search to work in other languages)

martinbedouret:
@shayc  I have added a couple of scripts to gather symbols and data from the ARASAAC API. Here's the complete package https://cboardgroupqadiag.blob.core.windows.net/arasaac/arasaac.zip

PR #1332, opened 2023-01-07
Feat/arasaac offline indexeddb

shayc:
POC
* can import arasaac.zip file (~289mb) to indexeddb via file input.
* after import, can search via text input

Screenshot 2023-01-08 at 1 23 07


RodriSanchez1:
close #1323

PR #1344, opened 2023-02-01
Feature/use premium required modal from app component

tomivm:
This PR prevent multiple mountings of PremiumRequiredModal by reducing its in a single mount on the App Component. It use redux actions to Show and hide the Modal


src/providers/SubscriptionProvider/SubscriptionProvider.reducer.js
```diff
@@ -61,6 +67,22 @@ function subscriptionProviderReducer(state = initialState, action) {
...
+          open: true,
+          isTryPeriodFinished: action.isTryPeriodFinished
+        }
+      };
+    case HIDE_PREMIUM_REQUIRED:
+      return {
+        ...state,
+        premiumRequiredModalState: {
+          open: false,
+          isTryPeriodFinished: false
```
RodriSanchez1: Why you do `isTryPeriodFinished: false` ?  I think this action need to change only the open state.

tomivm: The property should be renamed. this prop is to show the trial period ended messages. This is not to indicate the end of the trial period.

PR #1348, opened 2023-02-08
Feature/show trial period ended when open the app

tomivm:


src/providers/SubscriptionProvider/SubscriptionProvider.actions.js
```diff
@@ -12,6 +13,33 @@ import {
...
+      dispatch(showPremiumRequired({ showTryPeriodFinishedMessages: true }));
+
+    function isUserOnTrialPeriod(createdAt) {
+      const createdAtDate = new Date(createdAt);
+      const actualDate = new Date();
+      const DAYS_TO_TRY = 30;
+      const tryLimitDate = createdAtDate.setDate(
+        createdAtDate.getDate() + DAYS_TO_TRY
+      );
+      if (!createdAt) return false; //this case are already created users
```
RodriSanchez1: Move this condition to line 30 to avoid unnecessary process throughout the function

tomivm: When declaring `const` variable in a function would be elevated the declaration to the top of the function by js hoisting. Is true that If you try to access to that variable before declaration results in an error. But I think that there are on the Temporal Dead Zone. See https://www.educative.io/answers/what-is-the-temporal-dead-zone-in-javascript 
I'm not really sure if executable code that returns a value to the variable is executed on the hoisting process. I.

PR #1349, opened 2023-02-09
Unlock premium features if the user isOnTrialPeriod

tomivm:


src/components/PremiumFeature/PremiumFeature.js
```diff
@@ -2,23 +2,28 @@ import React from 'react';
...
-      event.stopPropagation();
-      props.showPremiumRequired();
-    }
+    if (isSubscribed || isOnTrialPeriod) return;
+    event.stopPropagation();
+    showPremiumRequired();
   };
 
-  console.log(props.children);
+  console.log(children);
```
RodriSanchez1: We can remove this

PR #1399, opened 2023-03-03
Fix/cloud voices on apple

### LINKED ISSUE #1398: cloud voices not work on IOS Apple devices
User Agent of IOS blocks cloud voices play

tomivm:
#1398

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -32,7 +32,6 @@ export const ADSENSE_CLIENT = 'ca-pub-7162313874228987';
 
 export const ADD_SLOT_SETTINGS_TOP = '5250438005';
 
-export const IS_BROWSING_FROM_APPLE = (() => {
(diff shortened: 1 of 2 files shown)


src/constants.js
```diff
@@ -31,3 +31,8 @@ export const ADTEST_AVAILABLE =
 export const ADSENSE_CLIENT = 'ca-pub-7162313874228987';
 
 export const ADD_SLOT_SETTINGS_TOP = '5250438005';
+
+export const IS_BROWSING_FROM_APPLE = (() => {
```
RodriSanchez1: I think thath you can do directly like this:  `export const IS_BROWSING_FROM_APPLE =/iPad|iPhone|iPod|Mac/.test(navigator.userAgent)`

martinbedouret (approved):

PR #1406, opened 2023-03-08
Fix/Output bar on Ios

### LINKED ISSUE #1397: Tiles on Output Bar are not sliding
When a tile is pressed, the output bar fills but does not slide.

tomivm:
In order to create a global solution for all browsers and user agents. This fix implements a new way to auto-scroll to the output bar's last symbol. 
close #1397

### DIFF
--- a/src/components/Board/Output/SymbolOutput/Scroll/Scroll.js
+++ b/src/components/Board/Output/SymbolOutput/Scroll/Scroll.js
@@ -5,8 +5,6 @@ import { Scannable } from 'react-scannable';
 
 import './Scroll.css';
 
-const invertDir = dir => (dir === 'rtl' ? 'ltr' : 'rtl');
-
 export class Scroll extends PureComponent {
   static propTypes = {
     /**
@@ -23,20 +21,16 @@ export class Scroll extends PureComponent {
       ...other
     } = this.props;
 
-    const scrollStyle = { direction: invertDir(direction) };
-
     return (
-      <div className="Scroll" style={scrollStyle}>
-        <Scannable>
-          <div
-            className="Scroll__container"
-            style={{ ...style, direction }}
-            {...other}
-          >
-            {children}
-          </div>
-        </Scannable>
-      </div>
+      <Scannable>
+        <div
+          className="Scroll__container"
+          style={{ ...style, direction }}
+          {...other}
+        >
+          {children}
+        </div>
+      </Scannable>
     );
   }
 }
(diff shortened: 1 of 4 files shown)


src/components/Board/Output/SymbolOutput/SymbolOutput.js
```diff
@@ -54,6 +54,21 @@ class SymbolOutput extends PureComponent {
     symbols: []
   };
```
RodriSanchez1, line 56: Add The ` scrollToLastSymbol ` funtion on `componentDidMount()`

src/components/Board/Output/SymbolOutput/SymbolOutput.js
```diff
@@ -54,6 +54,21 @@ class SymbolOutput extends PureComponent {
...
 
+  componentDidUpdate(prevProps) {
+    const scrollToLastSymbol = () => {
+      const { symbols } = this.props;
+      if (prevProps.symbols.length < symbols.length) {
+        try {
```
RodriSanchez1: @martinbedouret What do you think about use try catch block here?

src/components/Board/Output/SymbolOutput/Scroll/Scroll.js
```diff
@@ -23,20 +21,16 @@ export class Scroll extends PureComponent {
       ...other
     } = this.props;
 
-    const scrollStyle = { direction: invertDir(direction) };
```
martinbedouret, line 26: invert the scroll direction is necessary in case of languages written from right to left (Arabic, Hebrew )

src/components/Board/Output/SymbolOutput/SymbolOutput.js
martinbedouret: maybe necessary for the case that element is not found?

martinbedouret (changes requested):

src/components/Board/Output/SymbolOutput/Scroll/Scroll.js
tomivm: This is already contemplated on the Scroll__container. These lines were a part of the old way used to auto-scroll to the last element.

src/components/Board/Output/SymbolOutput/SymbolOutput.js
tomivm: Yes. That was the intention.

PR #1426, opened 2023-03-24
Disable Azure app insights telemetry on development

### LINKED ISSUE #1421: Disable azure application insights if the app is running on localhost
This article could be useful https://stackoverflow.com/questions/49311265/application-insights-how-to-exclude-localhost

tomivm:
close #1421 
This only solves the problem on the web App
To fix this on cordova Apps we should develop a way to differentiate the build for the production from the debug

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -22,7 +22,7 @@ export const AZURE_VOICES_BASE_PATH_API =
 
 // AdSense constants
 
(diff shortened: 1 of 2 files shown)


martinbedouret (approved):

martinbedouret:
great job @tomivm

PR #1445, opened 2023-04-10
Fix/record audio on ios APP

tomivm:
Allow the recording of audio to the Cordova IOS app


tomivm:
This change is not necessary. only set swift-language-version >= 4.2 in Xcode AAC Cboard target

PR #1447, opened 2023-04-10
Subscription fixes 

martinbedouret:


src/providers/SubscriptionProvider/SubscriptionProvider.actions.js
```diff
@@ -65,6 +57,42 @@ export function updateIsOnTrialPeriod() {
...
+        dispatch(
+          updateSubscription({
+            isSubscribed
+          })
+        );
+      } else {
+        const userId = state.app.userData.id;
+        const { status } = await API.getSubscriber(userId);
+        isSubscribed =
+          status.toLowerCase() === ('active' || 'canceled') ? true : false;
```
RodriSanchez1: Here we must also add the `in_grace_period` state and we can use the constants defined on SubscriptionProvider.constants

src/components/Settings/Subscribe/Subscribe.container.js
```diff
@@ -224,10 +215,11 @@ const mapStateToProps = state => {
 const mapDispatchToProps = {
   updateSubscriberId,
   updateSubscription,
-  comprobeSubscription,
+  checkSubscription: checkSubscription,
```
RodriSanchez1: Leave just one `checkSubscription`

PR #1463, opened 2023-04-22
Show password feature

### LINKED ISSUE #1354: Toggle password visibility
On log in section add a button that hide and show the password. Its helpful for the users to know what they are writing. Here is an example: 

maxresdefault

arjunanand23:
Attempted to add feature where password could be shown/hidden. Currently works if another character is entered but does not take place immediately.

### DIFF
--- a/src/components/Account/Login/Login.component.js
+++ b/src/components/Account/Login/Login.component.js
@@ -17,6 +17,11 @@ import { login } from './Login.actions';
 import messages from './Login.messages';
 import './Login.css';
 
+import InputAdornment from '@material-ui/core/InputAdornment';
+import IconButton from '@material-ui/core/IconButton';
+import Visibility from '@material-ui/icons/Visibility';
+import VisibilityOff from '@material-ui/icons/VisibilityOff';
+
 export class Login extends Component {
   static propTypes = {
     intl: intlShape.isRequired,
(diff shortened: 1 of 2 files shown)


martinbedouret:
awesome @arjunanand23 !
@tomivm Could you please take a look?

src/components/Account/Login/Login.component.js
```diff
@@ -78,9 +90,19 @@ export class Login extends Component {
...
-                  type="password"
+                  type={showPassword ? "text" : "password"}
                   name="password"
-                  onChange={handleChange}
+                  onChange={handleChange('password')}
+                  InputProps={{
+                    endAdornment: <InputAdornment position="end">
+                      <IconButton
+                        aria-label="toggle password visibility"
+                        onClick={() => { showPassword = !showPassword; }}
```
shayc, line 100: this line is problematic, updating a variable instead of state

src/components/Account/Login/Login.component.js
```diff
@@ -49,6 +59,8 @@ export class Login extends Component {
 
     const isButtonDisabled = isLogging || !!loginStatus.success;
 
+    var showPassword = false;
```
tomivm, line 62: A state should be used for this. This is the reason why you need to type a new character to see the password.

tomivm (changes requested):
Use State instead of a variable

tomivm:
completed on #1700 PR

PR #1535, opened 2023-07-11
Feature/Clear Arasaac db

### LINKED ISSUE #1529: Slide switch enabled after cancel download Arasaac symbols
image

RodriSanchez1:
On this PR: 
- Generat a static URL in the case that the tile was selected from the idb and is no internet conecction.
- Two buttons for download and delete the indexed database were added on the Symbols section.  NOTE: The idb will not be deleted, just cleared.
- Improve the error handlers. This will prevent if exist an error while downloading the package or while processing. 
- A signal controller was added on the arasaac file request, to avoid memory leaks if the user goes back while the download is active. 

close #1529


src/components/Settings/Symbols/Symbols.component.js
```diff
@@ -74,26 +80,37 @@ class Symbols extends React.Component {
...
+                        : this.downloadArasaacSymbols
                     }
-                  />
+                  >
+                    {symbolsSettings.arasaacActive ? (
+                      <DeleteForeverIcon color="secondary" fontSize="large" />
+                    ) : (
+                      <GetAppIcon fontSize="large" />
+                    )}
+                  </IconButton>
```
tomivm: You could create 2 different IconButton To achieve this.

src/components/Settings/Symbols/Symbols.container.js
```diff
@@ -77,20 +80,38 @@ export class SymbolsContainer extends PureComponent {
     });
   };
 
-  handleCompleted = async file => {
+  handleDeleteDialogAcepted = async () => {
+    this.setState({
+      isDeleting: true
+    });
+
+    await clearArasaacDB();
```
tomivm: This could fail? It would be good to wrap it on a try-catch block.

src/components/Settings/Symbols/DeleteArasaacDialog.js
```diff
@@ -0,0 +1,70 @@
...
+
+function DeleteArasaacDialog(props) {
+  const { open, onClose, onDialogAcepted, intl, isDeleting } = props;
+
+  const handleDialogAccepted = () => {
+    onDialogAcepted();
+  };
+
+  const handleClose = () => {
+    onClose();
```
tomivm, line 31: could pass the prop directly without the handle function

src/components/Settings/Symbols/Symbols.messages.js
```diff
@@ -3,42 +3,55 @@ import { defineMessages } from 'react-intl';
...
-    defaultMessage: 'No internet connection',
+    defaultMessage: 'No internet connection'
   },
   noConnectionDialogDesc: {
     id: 'cboard.components.Settings.Symbols.noConnectionDialogDesc',
     defaultMessage:
-      'No internet connection. You need to be connected to the internet to download symbols.',
+      'No internet connection. You need to be connected to the internet to download symbols.'
   },
+  deletArasaac: {
```
martinbedouret: typo `deletArasaac` Instead of ` deleteArasaac`

src/components/Settings/Symbols/Symbols.messages.js
```diff
@@ -3,42 +3,55 @@ import { defineMessages } from 'react-intl';
...
+    id: 'cboard.components.Settings.Symbols.deletArasaac',
+    defaultMessage: 'Delete ARASAAC Symbols'
+  },
+  deletArasaacSecondary: {
+    id: 'cboard.components.Settings.Symbols.deletArasaacSecondary',
+    defaultMessage:
+      'Delete the package with all ARASAAC Symbols. You will need an Internet connection to use it.'
```
martinbedouret: Rephrase `Delete the package with all ARASAAC Symbols. After deletion, the symbols will be only available with an Internet connection. `

src/components/UI/Downloader/Downloader.js
```diff
@@ -33,55 +33,63 @@ const DownloadItem = ({
...
+    loaded: 0
   });
 
   useEffect(
     () => {
-      if (downloadInfo.completed) return;
+      if (downloadInfo.completed || processing !== '') return;
+      const controller = new AbortController();
       const options = {
+        signal: controller.signal,
```
martinbedouret, line 50: this is well implemented, anyway I don't see it is working. I just turn off the connection during the download but nothing happens.

src/components/Settings/Symbols/Symbols.component.js
```diff
@@ -74,26 +80,34 @@ class Symbols extends React.Component {
...
-                        : false
-                    }
-                  />
+                <ListItemSecondaryAction
+                  style={{
+                    visibility: arasaacDownload.started ? 'hidden' : 'visible'
+                  }}
+                >
+                  {symbolsSettings.arasaacActive ? (
+                    <IconButton
```
martinbedouret: replace by a standard button that says `delete`

src/components/Settings/Symbols/Symbols.component.js
```diff
@@ -74,26 +80,34 @@ class Symbols extends React.Component {
...
+                      label={intl.formatMessage(messages.delete)}
+                      onClick={this.deleteArasaacSymbols}
+                    >
+                      <DeleteForeverIcon fontSize="large" />
+                    </IconButton>
+                  ) : (
+                    <IconButton
```
martinbedouret: replace by a standard button that says `download`

martinbedouret (approved):

src/components/UI/Downloader/Downloader.js
RodriSanchez1: This works when you go back to the last history page while downloading. Without this implementation you will see a memory leak error.

martinbedouret (approved):

PR #1539, opened 2023-07-13
Fix / Get subscriber loop

RodriSanchez1:
On this PR:
- Remove onResume cordova event
- Refreshes the subscriber status when a premium feature is clicked and the last update time was more than 12 hours ago.

### DIFF
--- a/src/components/PremiumFeature/PremiumFeature.js
+++ b/src/components/PremiumFeature/PremiumFeature.js
@@ -8,7 +8,7 @@ import {
   showPremiumRequired
 } from '../../providers/SubscriptionProvider/SubscriptionProvider.actions';
 
-function isUpdateSubscriptionNeeded(lastUpdated) {
+function isUpdateSubscriberStatusNeeded(lastUpdated) {
   if (!lastUpdated) return true;
   const MAX_HOURS_DIFFERENCE = 12;
   const actualTime = new Date().getTime();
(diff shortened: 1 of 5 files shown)


src/components/PremiumFeature/PremiumFeature.js
```diff
@@ -1,15 +1,38 @@
...
   children,
   isOnTrialPeriod,
   isSubscribed,
   isInFreeCountry,
-  showPremiumRequired
+  showPremiumRequired,
+  lastUpdated
 }) {
   const captured = event => {
+    if (isUpdateSubscriptionNeeded(lastUpdated)) {
```
tomivm: Probably would be good rename this function by something like "isUpdateIsSubscribedNeeded" or "isUpdateSubscriberStatusNeeded"

tomivm:
Please Add a Description :)

src/providers/SubscriptionProvider/SubscriptionProvider.container.js
martinbedouret, line 1: 'isCordova' is defined but never used

src/providers/SubscriptionProvider/SubscriptionProvider.reducer.js
```diff
@@ -37,7 +37,8 @@ const initialState = {
       price: '',
       tag: ''
     }
-  ]
+  ],
+  lastUpdated: null
```
martinbedouret: use Unassigned instead of  null here

martinbedouret (changes requested):

martinbedouret (approved):

PR #1551, opened 2023-07-21
Fix/refactor get subscriber not found

### LINKED ISSUE #1532: Refactor frontend and back-end to avoid 404 error in case a subscriber is not found
We are returning 404 error when find for a subscriber that doesn't exist: 

```
GET https://api.app.cboard.io/subscriber/5ada0d00eded2d001029c1ff 404 (Not Found)
```
We want to replace it with a 200 OK response and a payload indicating that the subscriber was not found. 
```
200 OK
{   success: false ,
     userid: user._id,
     message: 'ERROR: Subscriber was not found. '
 }
```

RodriSanchez1:
On this PR:
* Check in `getSubscriber()` request if the response have the succes prop, if not or is false throw the response data
* Support old behaviour for Android  <1.31.0 versions

Also see on CboardAPI [PR  - Fix/Refactor get subscriber ](https://github.com/cboard-org/cboard-api/pull/315)


RodriSanchez1:
close #1532

PR #1553, opened 2023-07-22
Fix/redirect to login page after activate an account

### LINKED ISSUE #1342: Users should keep session after activate a new acount
Now the new users have to log in after creating their accounts. this could be confusing. Users can think that's are already logged in after the redirection to the home page.

tomivm:
In this PR the Activate container is prepared to keep the user logged in after activating an account. it needs an update on the API  => https://github.com/cboard-org/cboard-api/pull/308 
It should be tested on Android with the universal Link feature. 
Close #1342


martinbedouret (changes requested):
@tomivm please refactor this to a simpler strategy: just redirect the user to the login page once the activation is done successfully. No API changes required.

tomivm:
@martinbedouret Now any change on the API is needed and the user is not logged in automatically. Users are redirected to the login-signup view and the messages on the Activate component are ready to be translated

martinbedouret (approved):

PR #1561, opened 2023-08-10
Feature/Redesign exported pdf boards

RodriSanchez1:
On this PR:
- Avoid displaying the name of the board on PicSeePal export
- Contain text in the tile box
- Avoid displaying a gray square on empty tiles; a white box will be displayed instead

image


martinbedouret (changes requested):
@RodriSanchez1 I love this feature. 
It works great if the labels of the tiles are below. Anyway if the labels are above, the pdf is wrong, see the image: 
image

martinbedouret (approved):

PR #1595, opened 2023-10-18
Bump/nodejs v18

RodriSanchez1:
On this PR:
* Node version was updated to `v18.18.1`. The new version was pinned on docker file, `.nvmrc` and CI config.
* React-scripts package was updated to v5. This changes comes with an upgrade of Webpack to v5. To make all works properly, the [CRACO](https://github.com/dilanx/craco) package was added for config webpack avoiding to eject. 
NOTE: We are currently react-rewire (similiar than CRACO) but is not maintened anymore. Needs to migrate.
* Change option `--maxWorkers=50%` for `--runInBand` on CI config. This change was made because the Unit Test were hanging with no output response. Also increment the run speed of the test.
* Fix default export on `Import.constants.js` and `Export.helpers.js`
* Ignore url webpack resolution on `WelcomeScreen.css`
* Fix `openDb` import on  `arasaac.ts `
* Update `yarn.lock` file


martinbedouret:
@RodriSanchez1 Shouldn't we remove the `react-scripts` package?

martinbedouret:
@RodriSanchez1 Just installed dependencies and run and got the following error: 

image

RodriSanchez1:
> @RodriSanchez1 Shouldn't we remove the `react-scripts` package?

No, we still use it. But an update to `react-scripts` was needed to support Node v18.
I'm planning to migrate to Vite and then we can remove it.

> @RodriSanchez1 Just installed dependencies and run and got the following error:

Do you change your node version to v18.18.1 before build? On my pc is building well.

package.json
```diff
@@ -100,12 +102,14 @@
     "husky": "^1.1.4",
     "jest-mock-axios": "^3.2.0",
     "lint-staged": "^10.0.3",
+    "node-polyfill-webpack-plugin": "^2.0.1",
     "prettier": "1.15.3",
     "react-app-rewired": "^2.1.3",
-    "react-scripts": "^3.4.4",
+    "react-scripts": "5.0.1",
```
shayc, line 108: Adding the breaking changes:
* v4 to v5: https://github.com/facebook/create-react-app/blob/main/CHANGELOG.md#breaking-changes
* v3 to v4: https://github.com/facebook/create-react-app/blob/main/CHANGELOG-4.x.md#breaking-changes

package.json
```diff
@@ -17,6 +17,7 @@
   "private": false,
   "license": "GPL-3.0-only",
   "dependencies": {
+    "@babel/plugin-proposal-private-property-in-object": "^7.21.11",
```
shayc: should be under `devDependencies`

package.json
```diff
@@ -144,19 +148,19 @@
   },
   "scripts": {
     "analyze": "source-map-explorer build/static/js/**/*.js",
-    "start": "react-scripts start",
-    "build": "react-scripts build && sw-precache --config=sw-precache-config.js",
+    "start": "craco start",
+    "build": "craco build && sw-precache --config=sw-precache-config.js",
     "build-cordova-debug": "react-app-rewired build",
```
shayc, line 153: is `react-app-rewired` needed? can you use craco instead? (i'm not sure)

craco.config.js
```diff
@@ -0,0 +1,36 @@
...
+        '.mjs': ['.mjs', '.mts']
+      }
+    },
+
+    configure: {
+      ignoreWarnings: [
+        function ignoreSourcemapsloaderWarnings(warning) {
+          return (
+            warning.module &&
+            warning.module.resource.includes('node_modules') &&
```
shayc: I think you can do 
```js
warning.module?.resource.includes('node_modules') && warning.details?.includes('source-map-loader')
```

.circleci/config.yml
```diff
@@ -7,7 +7,7 @@ jobs:
   build:
     docker:
       # specify the version you desire here
-      - image:  cimg/node:16.18
+      - image: cimg/node:18.18.1
```
shayc, line 10: Latest Long-Term Support Version: [20.9.0 (includes npm 10.1.0)](https://nodejs.org/en/download) - no breaking changes from v18 (or v16) - [ref](https://www.itmagination.com/blog/introducing-node-js-20-new-features-updates-and-improvements) - maybe we can upgrade to that?

.nvmrc
```diff
@@ -1 +1 @@
-v16
+v18.18.1
```
shayc: I wouldn't use too specific version in nvmrc. maybe just `v18.18`, this is so if there are bug patch updates (like node v18.18.2) the developer can update.

package.json
RodriSanchez1: Yes we can remove it. I will do that in other PR

PR #1610, opened 2023-11-06
Transform receipt on purchase validator

RodriSanchez1:
On this PR:

- The receipt object retrieved from the cordova-purchase-plugin needs to be transformed to support the new version 13.8.6.  See the ccboard [PR](https://github.com/cboard-org/ccboard/pull/71)

### DIFF
--- a/src/providers/SubscriptionProvider/SubscriptionProvider.container.js
+++ b/src/providers/SubscriptionProvider/SubscriptionProvider.container.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 
 import API from '../../api';
-import { isAndroid } from '../../cordova-util';
+import { isAndroid, isIOS } from '../../cordova-util';
 
 import {
   updateIsInFreeCountry,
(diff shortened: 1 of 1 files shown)


src/providers/SubscriptionProvider/SubscriptionProvider.container.js
```diff
@@ -67,9 +67,12 @@ export class SubscriptionProvider extends Component {
   };
 
   configPurchaseValidator = () => {
+    const transformReceipt = this.transformReceipt;
```
tomivm: The only use of tranformReceipt is inside configPurchaseValidator function. Declare the function tranformReceipt inside of this scope and avoid the use of 'this'.

src/providers/SubscriptionProvider/SubscriptionProvider.container.js
```diff
@@ -67,9 +67,12 @@ export class SubscriptionProvider extends Component {
...
+
     window.CdvPurchase.store.validator = async function(receipt, callback) {
       try {
-        const transaction = receipt.transactions[0];
+        const transaction = transformReceipt(receipt);
```
tomivm: Please avoid changes on IOS platform and make the changes only for Android.

tomivm (changes requested):

martinbedouret (approved):

PR #1635, opened 2023-12-12
Improve phrase to be allowed only by logged in users in free countries

### LINKED ISSUE #1627: Prevent enable the 'Improve Phrase' feature when user is not logged
Show an alert for unlogged users in free countries trying to enable the Improve Phrase feature.
Also disabled it on Logout. 
Could be possible to only show A message requiring the login on the Space for messages of the feature

maxicapodacqua:
Fixes #1627 
WIP

## Screenshot
image

## TODO:
- [x] Test interaction between PremiumRequiredModal and new modal (will event.stopPropagation break it?)
- [x] Dev test as a logged-in user
- [x] Disable feature on logout


src/components/Account/Login/Login.actions.js
```diff
@@ -48,6 +49,10 @@ export function logout() {
       console.error(err);
     }
   return async dispatch => {
+    dispatch(updateNavigationSettings({ improvePhraseActive: false }));
+    try {
+      await API.updateSettings({ navigation: { improvePhraseActive: false } });
+    } catch (e) {}
```
maxicapodacqua: @tomivm Is this what you mean by:
> Also disabled it on Logout.

in the issue?

currently, it deactivates it, but when I login again, it goes back to the old status (as it was in the backend), I think this will affect users in a negative way, as they will need to activate this again on every login

tomivm: yes! Only change the state on the front end. Change the local state with the `updateNavigationSettings`. Please don't worry about updating the API. If you update the setting for the user when he logs in again the feature will be disabled. if not, it will be automatically enabled on the login.

tomivm:
Thanks, @maxicapodacqua great Job getting familiar with the Cboard code. 
Excuse me. I pushed some commits to use the same capturer for premium features to also require the user to log in. (this is because right now the only features that gonna require the login of the users in free countries are also premium features. and would be easier than to have 2 different wrappers).
Let me know if you have doubts or comments, please! 
Again thank you so much!

src/components/LoggedInFeature/LoggedInFeature.js
tomivm, line 1: Please remove this file which is not necessary now.

src/components/LoggedInFeature/LoginRequiredModal.messages.js
```diff
@@ -0,0 +1,17 @@
+import { defineMessages } from 'react-intl';
+
+export default defineMessages({
+  featureBlockedTitle: {
+    id: 'cboard.components.LoginRequiredModal.featureBlockedTitle',
+    defaultMessage: 'This feature is blocked'
```
tomivm: please replace this text with something similar than:
'Log in to use this feature'

src/components/LoggedInFeature/LoginRequiredModal.messages.js
```diff
@@ -0,0 +1,17 @@
...
+  },
+  featureBlockedText: {
+    id: 'cboard.components.LoginRequiredModal.featureBlockedText',
+    defaultMessage:
+      'Cboard disabled this feature. To continue using it please sign up or login'
```
tomivm: please replace this text with something similar than:
'This feature is disabled for unlogged users. To continue using it please sign up or login'

tomivm (changes requested):

Issue #1632, opened 2023-12-11
Remove react-app-rewired package


RodriSanchez1:
We can migrate the implementation of react-app-rewired to craco package.

jakubsimo:
Hello, I have this issue resolved, I will soon push my changes and create PR, can you assign this to me ?

PR #1636, opened 2023-12-12
Removing react-app-rewired package - Issue #1632

jakubsimo:
Removing react-app-rewired package - Issue #1632  - now only using craco instead, also updated part of README which was mentioning react-app-rewired, react-app-rewired package uninstalled.


RodriSanchez1:
See #1696

PR #1638, opened 2023-12-13
Fix/prevent infinite loop on pasting boards

tomivm:
This PR has the objective of stopping the recursive function used to paste boards in the case of the user pasting a board inside of itself or in one of its children. To prevent an infinite loop that stops the app and creates unnecessary boards that create a memory problem on redux persistor and pass the limit of boards returned by getting Boards req 
**Before**
[Screen Recording 2023-12-12 at 18.04.58.zip](https://github.com/cboard-org/cboard/files/13654564/Screen.Recording.2023-12-12.at.18.04.58.zip)


tomivm:
Better solution on #1640

PR #1640, opened 2023-12-15
Fix/fix infinite loop on paste boards recursively

### LINKED ISSUE #1616: Dissalow paste a board Inside of it self
Paste a board Inside of itself causes an infinite loop that creates multiple undesired boards and causes bugs on the app.

tomivm:
**_This PR_**
Avoid an infinite loop by pasting a board inside of itself or on a child board. 
Close #1616 

**Before this**
[Screen.Recording.2023-12-12.at.18.04.58.zip](https://github.com/cboard-org/cboard/files/13679768/Screen.Recording.2023-12-12.at.18.04.58.zip)

**_To Test_**

**A**
  1) Copy a board
  2) Open that board and paste the board inside.
  A new Board Should be created that is an exact copy of the board copied (Also the folders inside continue to be linked to the same folders of the copied board. This is an old behavior.)

**B**
 1) Copy a board that contains subfolders
 2) Open that board and paste the board inside of a subfolder.
  A new Board Should be created that is an exact copy of the board copied (Also the folders inside continue to be linked to the same folders of the copied board. This is an old behavior.)


src/components/Board/Board.container.js
```diff
@@ -1517,10 +1525,11 @@ export class BoardContainer extends Component {
     //return condition
     newBoard.tiles.forEach(async tile => {
       if (tile && tile.loadBoard && !tile.linkedBoard) {
+        if (folderTile.loadBoard === firstPastedFolderId) return;
```
RodriSanchez1: This condition will be always true such as the params `folderTile.loadBoard` and `firstPastedFolderId` are equals on the `pasteBoardsRecursively` function call.

RodriSanchez1 (approved):

PR #1653, opened 2023-12-27
Adding message for pictogram not found

### LINKED ISSUE #1650: Create the UI to show the user that's not possible to find a pictogram
Create the UI to show the user that's not possible to find a pictogram for the searched word. Also, it's possible to invite the user to use our Pictonizer (new product) showing a coming soon.
image

maxicapodacqua:
Fixes #1650

### DIFF
--- a/src/components/Board/SymbolSearch/SymbolSearch.component.js
+++ b/src/components/Board/SymbolSearch/SymbolSearch.component.js
@@ -64,6 +64,7 @@ export class SymbolSearch extends PureComponent {
     super(props);
     this.state = {
       openMirror: false,
+      noSuggestions: false,
       value: '',
       suggestions: [],
       skin: 'white',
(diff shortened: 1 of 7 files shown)


maxicapodacqua:
This is ready for review @tomivm 😄

src/components/Board/SymbolSearch/SymbolSearch.component.js
```diff
@@ -429,6 +450,9 @@ export class SymbolSearch extends PureComponent {
             options={this.state.symbolSets}
             onChange={this.handleChangeOption}
           />
+          {!this.state.isFetching &&
```
tomivm: It's probably good to use an object for the isFetching state to avoid some flash appearance of the symbolNotFound when there aren't suggestions for some of the requested symbols banks and there are suggestions for other selected ones. like `isFetching{ arasaac: true, globalSymbol: true }` You could use a spread operator when you want to change one of them and preserve the other like `{isFetching{ ...this.state.isFetching, globalSymbol: true }`

tomivm (changes requested):

src/components/Board/SymbolSearch/SymbolSearch.component.js
maxicapodacqua: Makes sense, I made the requested  changes, but instead of a object I used different state vars to avoid the usage of the spread operator

tomivm (approved):

Issue #1487, opened 2023-05-17
Option to show tile text on UPPERCASE,  lowercase or First letter capitalized


RodriSanchez1:

cl-fontana:
This pull request should add this feature: #1674

PR #1674, opened 2024-03-31
Added new setting to display that can force labels to be in lowercase or uppercase

cl-fontana:
Hello!
I added the possibility of forcing the labels to lowercase or uppercase only. The default behavior is to leave the labels unchanged.

### DIFF
--- a/src/components/Settings/Display/Display.component.js
+++ b/src/components/Settings/Display/Display.component.js
@@ -22,7 +22,10 @@ import {
   DISPLAY_SIZE_EXTRALARGE,
   LABEL_POSITION_ABOVE,
   LABEL_POSITION_BELOW,
-  LABEL_POSITION_HIDDEN
+  LABEL_POSITION_HIDDEN,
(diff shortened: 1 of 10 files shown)


cl-fontana:
This should fix the issue #1487

src/components/Settings/Display/Display.component.js
```diff
@@ -113,44 +135,14 @@ class Display extends React.Component {
...
-                : DISPLAY_SIZE_STANDARD
-            }
-          >
-            {this.props.intl.formatMessage(
-              name === 'labelPosition'
-                ? messages[LABEL_POSITION_ABOVE]
-                : messages[DISPLAY_SIZE_STANDARD]
-            )}
+          <MenuItem value={values[name][0]}>
+            {this.props.intl.formatMessage(messages[values[name][0]])}
```
tomivm, line 139: Use a constant for the Array position to avoid Magic Numbers

tomivm (commented):
Thanks for your Help!

PR #1692, opened 2024-05-17
Fix / Paste subfolders

### LINKED ISSUE #1053: Copy-paste folder with subfolders doesnt work properly
When you copy-paste a folder that has two or more folders inside, only one changes its load board property. 

EDIT: 
The subfolders are saved on the DB but the parent board does not update the load board of each tile properly.

RodriSanchez1:
This PR replaces the forEach loop with a for-await-of loop in the `BoardContainer` component. The `pasteBoardsRecursively` function is now awaited for each iteration, ensuring that each call completes before moving on to the next one. 

close #1053


tomivm (approved):
Very good!

PR #1700, opened 2024-05-24
toggle password visibility

### LINKED ISSUE #1354: Toggle password visibility

nmro184:
resolving #1354

### DIFF
--- a/src/components/Account/Login/Login.component.js
+++ b/src/components/Account/Login/Login.component.js
@@ -36,7 +36,7 @@ export class Login extends Component {
   state = {
     isLogging: false,
     loginStatus: {},
-    showPassword: false
+    isPasswordVisible: false
   };
 
   handleSubmit = values => {
@@ -53,12 +53,12 @@ export class Login extends Component {
   };
   togglePasswordVisibility = () => {
     this.setState(prevState => ({
-      showPassword: !prevState.showPassword
+      isPasswordVisible: !prevState.isPasswordVisible
     }));
   };
 
   render() {
-    const { isLogging, loginStatus, showPassword } = this.state;
+    const { isLogging, loginStatus, isPasswordVisible } = this.state;
     const {
       intl,
       isDialogOpen,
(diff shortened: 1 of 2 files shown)


src/components/Account/Login/Login.component.js
```diff
@@ -32,7 +35,8 @@ export class Login extends Component {
 
   state = {
     isLogging: false,
-    loginStatus: {}
+    loginStatus: {},
+    showPassword: false
```
shayc: nit: `showPassword` sounds like a fn name (an action), would you mind renaming to something more boolean in nature? perhaps `isPasswordVisible`

nmro184: yes I agree

src/components/UI/FormItems/TextField.js
```diff
@@ -3,24 +3,36 @@ import PropTypes from 'prop-types';
...
 
 const defaultProps = {
   className: '',
-  error: false
+  error: false,
+  endAdornment: null
 };
 
-const TextField = ({ className, error, ...props }) => (
+const TextField = ({ className, error, endAdornment, ...props }) => (
```
RodriSanchez1: You are sending the `InputProps` prop that contains the `endAdornment`. So the correct prop to include here its the `InputProps`

src/components/UI/FormItems/TextField.js
```diff
@@ -3,24 +3,36 @@ import PropTypes from 'prop-types';
...
-  ])
+  ]),
+  endAdornment: PropTypes.node
 };
 
 const defaultProps = {
   className: '',
-  error: false
+  error: false,
+  endAdornment: null
```
RodriSanchez1: Change this to `InputProps: null`. See  line 24 comment. Also update the `propTypes` above

src/components/UI/FormItems/TextField.js
```diff
@@ -3,24 +3,36 @@ import PropTypes from 'prop-types';
...
 };
 
-const TextField = ({ className, error, ...props }) => (
+const TextField = ({ className, error, endAdornment, ...props }) => (
   <FormControl className={className} error={!!error}>
-    <MUITextField error={!!error} {...props} />
+    <MUITextField
+      error={!!error}
+      {...props}
+      InputProps={{
```
RodriSanchez1: If you make the changes that i suggested use  `InputProps={InputProps}` instead.

RodriSanchez1 (changes requested):
I get this error on the console `TextField.js:43 Warning: Failed prop type: The prop children is marked as required in ForwardRef(InputAdornment), but its value is null.`

Let me know if you have some doubts!!

nmro184:
thanks for the comments , when I run it locally I didn't get that warning , committed the changes you suggested.

nmro184:
@RodriSanchez1 hey I see the changes requested tag is still present. Is there anything else that needs to be fixed?

RodriSanchez1 (approved):
Good work!!! Thanks!

PR #1703, opened 2024-05-28
Fix/create-remote-Board-on-update-LoadBoards

### LINKED ISSUE #1690: Root board updated just on local state
The root board is only updated in the local state. This occurs when the user makes a modification to a subfolder and the user's root board is not saved in the database. The issue appears when the user logs in on another device or will appear if we fix issue #1689.

How to solve it: 
Create the user root board on the db when a subfolder is modified

tomivm:
Close #1690 
Posting every default board that uses a newly created API board id as loadBoard. To avoid losing boards between devices and logouts.
This ensures navigations between folders are using the loadBoard IDs and avoids navigating comparing the folder label and the Board Names. See https://github.com/tomivm/cboard/blob/d9943f82c3da4bb7e45f9c821588bebfe0b1569e/src/components/Board/Board.container.js#L888
This refactors pieces of two of our main functions. Please do a strong test.
Check how boards are updated and created.
Take care that the data on the Redux store and the data on the DB should be equal. Also, check if all the loadBoards IDs that should point to a remote board are correctly updated on our DB and in the local storage.


src/components/Board/Board.actions.js
```diff
@@ -674,10 +676,10 @@ export function updateApiObjectsNoChild(
   createCommunicator = false,
   createParentBoard = false
 ) {
-  return (dispatch, getState) => {
+  return async (dispatch, getState) => {
     //create - update parent board
     const action = createParentBoard ? createApiBoard : updateApiBoard;
-    return dispatch(action(parentBoard, parentBoard.id))
+    return await dispatch(action(parentBoard, parentBoard.id))
```
RodriSanchez1, line 704: When you use await, it pauses the execution of the function until the awaited promise resolves. Similarly, when you use .then(), it chains onto the completion of the promise and waits for it to resolve. Using both together is redundant because both achieve the same outcome of waiting for the completion of the asynchronous operation.
From Chat GPT

src/components/Board/Board.actions.js
```diff
@@ -722,27 +724,68 @@ export function updateApiObjectsNoChild(
   };
 }
 export function updateApiMarkedBoards() {
-  return (dispatch, getState) => {
+  return async (dispatch, getState) => {
     const allBoards = [...getState().board.boards];
-    for (let i = 0; i < allBoards.length; i++) {
+    for (const board of allBoards) {
```
RodriSanchez1: use `for await (const board of allBoards)`

src/components/Board/Board.constants.js
```diff
@@ -41,3 +43,4 @@ export const DOWNLOAD_IMAGE_SUCCESS = 'cboard/Board/DOWNLOAD_IMAGE_SUCCESS';
 export const DOWNLOAD_IMAGE_FAILURE = 'cboard/Board/DOWNLOAD_IMAGE_FAILURE';
 export const DEFAULT_ROWS_NUMBER = 5;
 export const DEFAULT_COLUMNS_NUMBER = 5;
+export const REMOTE_BOARD_ID_LENGTH = 14;
```
RodriSanchez1: The mongo object id has a 28 characters length. The max length of the random id generated by the shortid library is 14.

src/components/Board/Board.container.js
```diff
@@ -535,7 +535,11 @@ export class BoardContainer extends Component {
           updateBoard(boardData);
         }
         //api updates
-        updateApiObjectsNoChild(boardData, createCommunicator, createBoard)
+        await updateApiObjectsNoChild(
```
RodriSanchez1: If we want to await the `updateApiObjectNoChild` function, we should remove the promise chain. 
Something like: 
```js
    try{
        const boardId = await updateApiObjectsNoChild(
          boardData,
          createCommunicator,
          createBoard
        )
        if (createBoard) {
            replaceBoard({ ...boardData }, { ...boardData, id: boardId });
        }
        this.props.history.replace(`/board/${boardId}`);
    }
    catch(e){
        console.error(e);
    }
```

RodriSanchez1: I the suggested solution is implemente, we could remove the extenal try catch block.

PR #1705, opened 2024-05-30
added security step before account deletion

### LINKED ISSUE #1699: Add security step before delete user account
When a user wants to delete their account, after pressing the delete account button, a confirmation modal is displayed without any security measures. This lack of security can lead to unintentional deletions, such as a child accidentally accessing and deleting the account, resulting in the loss of all the user's data.

To prevent this, implementing a security step before deletion is essential. I suggest asking the user to write a confirmation word, similar to the method used by Mongo Compass.

The word can be: cboard, delete, delete-account, or any that the dev want. 

image

nmro184:
resolving #1699

### DIFF
--- a/src/components/Settings/People/DeleteConfirmationDialog.js
+++ b/src/components/Settings/People/DeleteConfirmationDialog.js
@@ -1,6 +1,6 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
-
+import { TextField } from '../../UI/FormItems';
 import { FormattedMessage } from 'react-intl';
 import messages from './People.messages';
 
@@ -31,6 +31,34 @@ const DeleteConfirmationDialog = ({
   isDeletingAccount,
   errorDeletingAccount
 }) => {
+  const [requiredConfirmationText, setRequiredConfirmationText] = useState({});
+  const [confirmationText, setConfirmationText] = useState('');
+  const [showConfirmationInput, setShowConfirmationInput] = useState(false);
+
+  useEffect(() => {
+    setRequiredConfirmationText(
+      <FormattedMessage {...messages.confirmationText} />
+    );
+  }, []);
+
+  useEffect(
+    () => {
+      if (!open) {
+        setConfirmationText('');
+        setShowConfirmationInput(false);
+      }
+    },
+    [open]
+  );
+
+  const handleConfirmationChange = e => {
+    setConfirmationText(e.target.value);
+  };
+
+  const handleShowConfirmationInput = () => {
+    setShowConfirmationInput(true);
+  };
+
   return (
     <Dialog
       open={open}
@@ -52,13 +80,13 @@ const DeleteConfirmationDialog = ({
           </DialogContentText>
         )}
       </DialogContent>
-      {!isDeletingAccount && (
+      {!isDeletingAccount && !showConfirmationInput && (
         <DialogActions>
           <Button
             variant="outlined"
             color="secondary"
             className={'delete_button'}
-            onClick={handleDeleteConfirmed}
+            onClick={handleShowConfirmationInput}
           >
             {<FormattedMessage {...messages.deleteAccountPrimary} />}
           </Button>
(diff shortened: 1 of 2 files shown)


src/components/Settings/People/DeleteConfirmationDialog.js
```diff
@@ -1,6 +1,6 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
-
+import { TextField } from '../../UI/FormItems';
```
RodriSanchez1: Use the `TextField` component from material UI.
`import {TextField} from from '@material-ui/core';`

src/components/Settings/People/DeleteConfirmationDialog.js
```diff
@@ -67,6 +95,34 @@ const DeleteConfirmationDialog = ({
           </Button>
         </DialogActions>
       )}
+      {!isDeletingAccount && showConfirmationInput && (
+        <>
+          <TextField
```
RodriSanchez1: This component needs to be inside the `<DialogContent>`

src/components/Settings/People/People.messages.js
```diff
@@ -54,6 +54,10 @@ export default defineMessages({
     defaultMessage:
       'Once you delete an account, there is no going back. Please be certain.'
   },
+  deleteAccountFinal: {
+    id: 'cboard.components.Settings.People.deleteAccountFinal',
+    defaultMessage: 'To confirm, type "delete-account" in the box below.'
```
RodriSanchez1: Chage the text to: "To confirm, type "delete-account"." See the Changes Requested comment conclusion

src/components/Settings/People/DeleteConfirmationDialog.js
```diff
@@ -31,6 +31,34 @@ const DeleteConfirmationDialog = ({
   isDeletingAccount,
   errorDeletingAccount
 }) => {
+  const [requiredConfirmationText, setRequiredConfirmationText] = useState({});
+  const [confirmationText, setConfirmationText] = useState('');
+  const [showConfirmationInput, setShowConfirmationInput] = useState(false);
+
+  useEffect(() => {
```
RodriSanchez1: This useEffect is not neccesary. I will leave on conclusion how you can deal with this.

src/components/Settings/People/People.messages.js
```diff
@@ -67,5 +71,9 @@ export default defineMessages({
   cancelDeleteAccount: {
     id: 'cboard.components.Settings.People.cancelDeleteAccount',
     defaultMessage: 'Cancel'
+  },
+  confirmationText: {
```
RodriSanchez1: This will be not neccesary.

RodriSanchez1 (changes requested):
@nmro184 good work! I will leave some comments: 

image
This needs to be our dialog when it opens. The `<TextField>` should be displayed directly, eliminating the intermediate step of clicking the "Delete Account" button to reveal the text field.

Additionally, create a constant  `const DELETE_ACCOUNT='delete-account'` and use it to compare the entered text and also to display it as a part of the label message. This approach allows you to remove both useEffect hooks.

To use a constant in the Formated Message: 

```js
DeleteConfirmationDialog.js 

          <TextField
            autoFocus={true}
            label={
              <FormattedMessage
                {...messages.deleteAccountFinal}
                values={{ deleteAccount: 'delete-account' }}
              />
            }
           ......../>

People.messages.js

 deleteAccountFinal: {
    id: 'cboard.components.Settings.People.deleteAccountFinal',
    defaultMessage: 'To confirm, type "{deleteAccount}".',
  },
```

One more thing, ensure the `confirmationText` value is cleared when the dialog is closed.

src/components/Settings/People/People.messages.js
nmro184: deleted

src/components/Settings/People/DeleteConfirmationDialog.js
nmro184: I used added a function called handleCloseDialog that resets the confirmation text

src/components/Settings/People/People.messages.js
nmro184: done

src/components/Settings/People/DeleteConfirmationDialog.js
nmro184: here I had a problem when I put the text field element inside DialogContent it shrinks and sticks to the left , also if its inside DialogActions , or outside of DialogContent in general it sticks to both sides of the dialog which doesn't look good , I know using inline css is not a best practice but I didn't know whether to create a new css file and if so where should I put it .

nmro184: also made sense to me to put it inside DialogActions because its kinda serves a role of a button , you know what I mean ? anyway its ok Im here to learn :)

RodriSanchez1: I see!! Take a look on material ui documentation under form dialog [here](https://v4.mui.com/components/dialogs/). The text field it's inside the dialog content. Use that as a reference. Also you can take a look on the text field props. 
You can add the fullWidth  prop to deal with your problem!

The css file is not needed

nmro184: oh got it , I have committed the changes @RodriSanchez1

PR #1707, opened 2024-06-07
Refactor / Communicator creation and updates 

### LINKED ISSUE #1418: Defaults boards are not saved as home if is selected from the tour dialog
The first time that the user logs in, a joyride tour is showed showing the different root boards that the user can have. If the user press the picseepal as root board, the active communicator is not updated and the root board is not saved

### LINKED ISSUE #1701: Implement the upsertApiCommunicator action

RodriSanchez1:
On this PR:
* A new function `verifyAndUpsertCommunicator()` was created. This function checks if the logged user has his own communicator, and if is not creates one. After that, the communicator is upserted.  Note that this function is just for upsert the local store.
* Implement the `upsertApiCommunicator()` when is necessary

close #1701
close #1418

### DIFF
--- a/src/components/Board/Board.actions.js
+++ b/src/components/Board/Board.actions.js
@@ -44,8 +44,6 @@ import {
 import API from '../../api';
 
 import {
-  updateApiCommunicator,
-  createApiCommunicator,
   replaceBoardCommunicator,
   upsertCommunicator,
   getApiMyCommunicators,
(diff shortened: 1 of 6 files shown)


tomivm (approved):
Good job!! it's more legible and easy to understand now. We have to be careful while merging the rest of the PR because of the modifications in updateApiObjects and updateApiObjectsNoChild. We have to do an exhaustive test to avoid unexpected behaviors 🚀

PR #1711, opened 2024-06-14
Feature/add only necessary default boards on login

### LINKED ISSUE #1710: Add just the necessary default boards on user login
Avoid adding unnecessary boards to the board.boards array on the store when the user is logged in.

tomivm:
This PR has the objective of including only the necessary default boards on the user's login.
Also, have changes to be sure that these boards would be added if the user requests it and aren't correctly added.
Please take a strong test!
close #1710


src/components/Board/Board.container.js
```diff
@@ -868,8 +876,22 @@ export class BoardContainer extends Component {
...
+        const findBoardOnStore = boardId =>
+          this.props.boards.find(b => b.id === boardId);
+
+        const nextBoard = findBoardOnStore(loadBoardSearched);
+        if (nextBoard) return nextBoard;
+        if (
+          ALL_DEFAULT_BOARDS.map(({ id }) => id).includes(loadBoardSearched)
+        ) {
+          addNecessaryDefaultBoardsFor(loadBoardSearched);
+          const nextBoard = findBoardOnStore(loadBoardFinder);
```
tomivm: Fix this. findBoardOnStore should receive a id

tomivm:
@RodriSanchez1 Please check the new changes. I improved the way to remove the unnecessary default boards. This changes on the API are needed https://github.com/cboard-org/cboard-api/pull/360

src/components/Communicator/Communicator.actions.js
```diff
@@ -291,3 +300,34 @@ export function updateDefaultBoardsIncluded(boardAlreadyIncludedData) {
...
+        return updatedCommunicatorData?.defaultBoardBlackList.concat(boardId);
+      return updatedCommunicatorData?.defaultBoardBlackList;
+    };
+
+    updatedCommunicatorData.defaultBoardBlackList = updatedCommunicatorData?.defaultBoardBlackList
+      ? concatBoardIdIfNecessary()
+      : [boardId];
+
+    dispatch(verifyAndUpsertCommunicator(updatedCommunicatorData));
+    return dispatch(upsertApiCommunicator(updatedCommunicatorData))
```
RodriSanchez1: This async action dispatch can be avoided such as the communicator gets updated later on. Just update the local state and be sure to get the updated state before the call of the `upsertApiCommunicator` action

PR #1718, opened 2024-06-24
Feature/edit folder load board

### LINKED ISSUE #1719: Add the capability to find losted folders
Users who lost folders. He should have a way to fix his boards by searching over his boards

tomivm:
This feature adds the capability to change loadBoards for the tile folders. This could really help our users that have wrong loadsBoards on their folders caused by errors on copying public boards and copying and pasting boards.
this is how it looks if the load folder matches with one of his folders:
image
this is how it looks if there is an error and the folder is lost.
image
Inside you can see ALL REMOTE boards and select One confirming on the dialog
image
image
Also, you can search for it by the name on the APP bar
image
Close #1719


tomivm (commented):
@RodriSanchez1 Can you review this please! it have long time in a backstage and it is a good feature to be implemented to our users that have errors in his boards

tomivm:
Probably we should remove the logic to warning loadBoards pointing to local boards until work more in it and explain users how to sync boards and merge this with less changes.

src/components/Board/Board.messages.js
```diff
@@ -72,7 +72,7 @@ export default defineMessages({
   boardMissed: {
     id: 'cboard.components.Board.boardMissed',
     defaultMessage:
-      'We are sorry but we have missed this folder / board. We recommend you to create it again.'
+      'We are sorry but we have missed this folder / board. Edit this tile to find it again.'
```
RodriSanchez1: What do you think if we use something like: 

`Sorry, this folder or board is missing. Edit this tile to link it again.`

src/components/Board/TileEditor/TileEditor.component.js
```diff
@@ -482,7 +497,7 @@ export class TileEditor extends Component {
     );
 
     const selectBoardElement = (
-      <div>
+      <div style={{ marginTop: '16px' }}>
```
RodriSanchez1, line 501: why dont you add this inline style to the css file?

src/components/Board/TileEditor/TileEditor.component.js
```diff
@@ -441,6 +448,14 @@ export class TileEditor extends Component {
     }
   };
 
+  handleLoadBoardChange = ({ boardId }) => {
+    if (boardId) {
+      this.props.onAddApiBoard(boardId);
+      this.updateTileProperty('loadBoard', boardId);
+      this.setLinkedBoard();
```
RodriSanchez1: Is neccesary to set the linked board prop when the load board changes?

src/components/Board/TileEditor/TileEditor.component.js
```diff
@@ -654,6 +680,34 @@ export class TileEditor extends Component {
...
                       selectBoardElement}
+
+                    {haveLoadBoard &&
+                      !isLocalLoadBoard &&
+                      !isCordova() &&
+                      this.editingTile() && (
+                        <>
+                          <FormLabel
+                            id="boards-input-label"
+                            style={{ marginTop: '16px' }}
```
RodriSanchez1, line 692: Same here, could we add this to the css file?

src/components/Board/TileEditor/TileEditor.messages.js
```diff
@@ -68,5 +68,28 @@ export default defineMessages({
...
+  loadBoardAlertDescription: {
+    id: 'cboard.components.Board.TileEditor.loadBoardAlertDescription',
+    defaultMessage:
+      'Try to find it manualy on your remote folders by clicking on the search button.'
+  },
+  loadBoardAlertDescriptionLocalId: {
+    id: 'cboard.components.Board.TileEditor.loadBoardAlertDescriptionLocalId',
+    defaultMessage: `It's looks like this folder is localy stored on the device that you
+          create it. If you want to use it, please make a change in it connected
+          to the internet. Or edit this value to use another folder.`
```
RodriSanchez1: I suggest to change to: 
_It looks like this folder is stored locally on the device where it was created. To use it, please connect that device to the internet and make a change to the folder. Alternatively, edit this tile to select a different folder._

src/components/Board/TileEditor/TileEditor.component.js
```diff
@@ -441,6 +448,14 @@ export class TileEditor extends Component {
     }
   };
 
+  handleLoadBoardChange = ({ boardId }) => {
+    if (boardId) {
+      this.props.onAddApiBoard(boardId);
```
RodriSanchez1, line 453: this is an async function. We should await for it or doesnt matter?

src/components/Board/TileEditor/LoadBoardEditor/LoadBoardEditor.js
```diff
@@ -0,0 +1,368 @@
...
+          </Toolbar>
+        </AppBar>
+        <div className={styles.boardsListContainer}>
+          {!loading && !error && (
+            <BoardPagination
+              handleChange={handleChangeOnPage}
+              pagesCount={totalPages}
+              currentPage={currentPage}
+            />
+          )}
```
RodriSanchez1, line 292: I suggest to avoid the double pagination rendering. With the bottom one is sufficient. 
It's quite extrange. It's on you

src/components/Board/TileEditor/LoadBoardEditor/LoadBoardEditor.js
RodriSanchez1, line 1: The scroll bar is placed for all the page. Take a  look 
image

Probably we should avoid adding it into the the AppBar

src/components/Board/TileEditor/LoadBoardEditor/LoadBoardEditor.messages.js
```diff
@@ -0,0 +1,46 @@
+import { defineMessages } from 'react-intl';
+
+export default defineMessages({
+  confirmationTitle: {
+    id: 'cboard.components.LoadBoardEditor.confirmationTitle',
+    defaultMessage:
+      'Are you sure you want change the board to open by this tile?'
```
RodriSanchez1: It can be: 

Are you sure you want to change the board that opens from this tile?

or

Are you sure you want to change which board this tile opens?

RodriSanchez1:
It looks good! I realley liked. I'm thinking if there is way to replace the previous link to an existing board with this new feature. What do you thing? This is possible? If not if kind of confusing. See the following image

image

src/components/Board/Board.messages.js
tomivm: Fine!

src/components/Board/TileEditor/LoadBoardEditor/LoadBoardEditor.js
tomivm: Are okey for me, we can remove it later

src/components/Board/TileEditor/LoadBoardEditor/LoadBoardEditor.js
tomivm: I will improve this in a next step. I opened an issue

src/components/Board/TileEditor/TileEditor.component.js
tomivm: Could be good to add a loading state but is not strictly necessary because this handler dont mount the board in it uses. We should add logic to manage error cases

src/components/Board/TileEditor/TileEditor.component.js
tomivm: it was for syncing the linked board and the linked folder

tomivm: I improved the function to avoid need the state to use it

tomivm:
@RodriSanchez1 Yes the plan is to unifique the link board and link folder ! It's not ready to do it now because  this feature was developed to be online only and the link board works offline and with local boards also. 
Hope we can improve this soon!
About your concerns: I agree, is strange have the both fields with the same value. For that I make a change to only show the corresponding value depending on if you use one board of the communicator or a folder!
Thanks for your review ! 🚀

PR #1725, opened 2024-07-02
Feature/sync communicator

RodriSanchez1:
For the test, you can log in on two different devices and make changes on each one. After the board is mounted you will have the latest communicator version on the device.

We need to think about how we will deal with the creation of two users communicators. This can produce issues

See this [cboard-api PR](https://github.com/cboard-org/cboard-api/pull/361)


src/components/Communicator/Communicator.actions.js
```diff
@@ -291,3 +298,84 @@ export function updateDefaultBoardsIncluded(boardAlreadyIncludedData) {
...
+    }
+
+    const activeCommunicatorId = getActiveCommunicator(getState).id ?? null;
+    const lastRemoteSavedCommunicatorId = remoteCommunicators[0].id ?? null; //The last communicator saved on the server
+    const needToChangeActiveCommunicator =
+      activeCommunicatorId !== lastRemoteSavedCommunicatorId &&
+      updatedCommunicators.length &&
+      updatedCommunicators.findIndex(
+        communicator => communicator.id === lastRemoteSavedCommunicatorId
+      ) !== -1;
```
tomivm, line 369: also we could check if `lastRemoteSavedCommunicatorId` is not null.

src/components/Communicator/Communicator.actions.js
```diff
@@ -291,3 +298,84 @@ export function updateDefaultBoardsIncluded(boardAlreadyIncludedData) {
...
+
+    dispatch({
+      type: SYNC_COMMUNICATORS,
+      communicators: updatedCommunicators,
+      activeCommunicatorId: needToChangeActiveCommunicator
+        ? remoteCommunicators[0].id
```
tomivm: is posible to use `lastRemoteSavedCommunicatorId` to make it more redeable.

src/components/Communicator/Communicator.actions.js
```diff
@@ -230,12 +234,15 @@ export function verifyAndUpsertCommunicator(
...
+  return async dispatch => {
     dispatch(getApiMyCommunicatorsStarted());
     return API.getCommunicators()
       .then(res => {
         dispatch(getApiMyCommunicatorsSuccess(res));
-        return res;
+
+        return dispatch(syncCommunicators(res.data)).catch(e => {
+          console.error(e);
+        });
```
tomivm: would be better to return the `res` after the dispatch of `syncCommunicators`. to be more consistence with the old behavior.

src/components/Account/Login/Login.actions.js
```diff
@@ -147,7 +147,8 @@ export function login({ email, password, activatedData }, type = 'local') {
       );
 
       if (loginData.communicators && loginData.communicators.length) {
-        currentCommunicator = loginData.communicators[0];
+        currentCommunicator =
+          loginData.communicators[loginData.communicators.length - 1]; //use the latest communicator
```
tomivm: would be good to create a const for `loginData.communicators.length - 1` to avoid magic position

src/components/Communicator/Communicator.actions.js
RodriSanchez1: done

src/components/Communicator/Communicator.actions.js
RodriSanchez1: done

src/components/Account/Login/Login.actions.js
RodriSanchez1: done

PR #1727, opened 2024-07-04
Fix / Unexpected behaviour on change default boards included

### LINKED ISSUE #1720: New root board is created when change the featured board
This occurs when a user creates a new API communicator, logs in on another device, and changes the featured board. If the new communicator was created in the database without a  `defaultIncludedBoards` value,  that would be an empty array.  

At the time of changing the featured board,  we do this comprobation:
```javascript
const defaultBoardsIncluded =
      activeCommunicator.defaultBoardsIncluded  //here we have a wrong truthy value 
      ||  fallbackInitialDefaultBoardsIncluded(activeCommunicator);
```
This causes the loss of the current root board on the user communicator

To be able to fix this bug, we need to add an exhaustive comprobation of the `defaultIncludedBoards`  prop.

Suggestions: 
```javascript
const isValid = (activeCommunicator.defaultBoardsIncluded && activeCommunicator.defaultBoardsIncluded.length > 0) ? true : false;
//or
const isValid = !!(activeCommunicator.defaultBoardsIncluded && activeCommunicator.defaultBoardsIncluded.length);
```

### LINKED ISSUE #1721: Create new communicator with correct values for defaultsBoardsIncluded

RodriSanchez1:
close #1720 
close #1721

### DIFF
--- a/src/api/communicators.json
+++ b/src/api/communicators.json
@@ -7,6 +7,10 @@
     "email": "support@cboard.io",
     "rootBoard": "root",
     "boards": ["root"],
-    "defaultBoardBlackList": []
+    "defaultBoardBlackList": [],
+    "defaultBoardsIncluded": [
+      { "nameOnJSON": "advanced", "homeBoard": "root" },
+      { "nameOnJSON": "picSeePal", "homeBoard": "jjmlUcQs19" }
+    ]
   }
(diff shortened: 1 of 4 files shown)


tomivm:
Awesome!

PR #1728, opened 2024-07-04
Fix/saveUserDataOnly

RodriSanchez1:


tomivm:
This could be closed if we merge #1731 and this PR in the API https://github.com/cboard-org/cboard-api/pull/362

PR #1743, opened 2024-07-23
Fix / Create neccesary default boards on DB

RodriSanchez1:
Comprobe if an API board contains the tile ID of a default board before marking it to create. Maybe with this solution we can avoid the creation of the defaultBoardBlackList


tomivm (approved):
Awesome! it's a better solution than using a blacklist. I agree that with these changes we can avoid the creation of the defaultBoardBlackList.

PR #1819, opened 2025-03-21
bug: navigationHistory

### LINKED ISSUE #1745: toRootBoard function is not updating the URL
#### **Description:**  
When the `toRootBoard` action is dispatched, the active board correctly updates to the root board. However, the URL still displays the previous board ID instead of reflecting the root board.  

#### **Steps to Reproduce:**  
1. Enable context-aware back buttons:  
   - Go to **Settings > Navigation and Buttons > Enable Context-Aware Back Buttons**  
   - Click **Save**  
2. Navigate to a deeply nested folder (at least three levels deep).  
3. Press the **|<** button to navigate to the root board.  
4. Observe that while navigation to the root board occurs, the URL does not update accordingly.  

#### **Expected Behavior:**  
The URL should update to reflect the root board when the `toRootBoard` action is triggered.  

#### **Actual Behavior:**  
The board changes, but the URL remains unchanged, displaying the previous board ID.  

#### **Additional Notes:**  
- This issue affects usability, as the URL does not accurately represent the user's current location.  
- It may cause inconsistencies when users refresh the page or share links.

gonojuarez:
add the follow code:
```js
 onRequestToRootBoard() {
    this.props.toRootBoard();
    const prevBoardId = this.props.navHistory[0];
    this.props.history.replace(`/board/${prevBoardId}`);
    this.scrollToTop();
  }
```

close #1745

### DIFF
--- a/src/components/Board/Board.actions.js
+++ b/src/components/Board/Board.actions.js
@@ -89,6 +89,12 @@ function getActiveCommunicator(getState) {
   return activeCommunicator;
 }
 
+function replaceHistory(getState) {
+  const { board } = getState();
+  const id = board.activeBoardId;
+  history.replace(`/board/${id}`);
+}
+
 export function changeDefaultBoard(selectedBoardNameOnJson) {
   return (dispatch, getState) => {
     const board = DEFAULT_BOARDS[selectedBoardNameOnJson];
@@ -266,15 +272,20 @@ export function changeBoard(boardId) {
 }
 
 export function previousBoard() {
-  return {
-    type: PREVIOUS_BOARD
+  return (dispatch, getState) => {
+    dispatch({
+      type: PREVIOUS_BOARD
+    });
+    replaceHistory(getState);
   };
 }
 
 export function toRootBoard() {
-  return dispatch => {
-    history.replace('/board/root');
-    dispatch({ type: TO_ROOT_BOARD });
+  return (dispatch, getState) => {
+    dispatch({
+      type: TO_ROOT_BOARD
+    });
+    replaceHistory(getState);
   };
 }
 
--- a/src/components/Board/Board.actions.js
+++ b/src/components/Board/Board.actions.js
@@ -89,7 +89,7 @@ function getActiveCommunicator(getState) {
   return activeCommunicator;
 }
 
-function replaceHistory(getState) {
+function replaceHistoryWithActiveBoardId(getState) {
   const { board } = getState();
   const id = board.activeBoardId;
   history.replace(`/board/${id}`);
@@ -276,7 +276,7 @@ export function previousBoard() {
     dispatch({
       type: PREVIOUS_BOARD
     });
-    replaceHistory(getState);
+    replaceHistoryWithActiveBoardId(getState);
   };
 }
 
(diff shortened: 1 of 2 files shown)


RodriSanchez1 (changes requested):
Gonza, your solution works great! Just one suggestion: we should move this logic into the `Board.actions` itself, so that when the action is dispatched, it also updates the history to point to the root board ID. That way, the route change stays consistent with the state change, and the responsibility is encapsulated within the action.

```js
export function toRootBoard() {
  return (dispatch) => {
    history.replace('/boards/root');
    dispatch({ type: TO_ROOT_BOARD });
  };
}
```

src/components/Board/Board.actions.js
```diff
@@ -272,8 +272,9 @@ export function previousBoard() {
 }
 
 export function toRootBoard() {
-  return {
-    type: TO_ROOT_BOARD
+  return dispatch => {
+    history.replace('/board/root');
```
RodriSanchez1: It should not redirect to `/board/root` , which is the default board. Instead, it should redirect to the user's root board.

RodriSanchez1 (changes requested):

src/components/Board/Board.container.js
```diff
@@ -1185,14 +1185,18 @@ export class BoardContainer extends Component {
     }
   };
 
+  propsHistoryReplace(boardId) {
```
tomivm: We can improve the name for this method. If it intention is replace `boardId` in the history. The name could be something like `historyReplaceBoardId``

src/components/Board/Board.container.js
```diff
@@ -1185,14 +1185,18 @@ export class BoardContainer extends Component {
...
+    this.props.previousBoard();
+    /*if (this.props.navHistory.length >= 2) {
       const prevBoardId = this.props.navHistory[
         this.props.navHistory.length - 2
       ];
-      this.props.history.replace(`/board/${prevBoardId}`);
-      this.scrollToTop();
-    }
+      this.propsHistoryReplace(prevBoardId);*/
+    this.scrollToTop();
```
tomivm, line 1194: This change look good. Please avoid leaving comments with old lines in the code. Make an atomic commit with a good title to documment what do you made in the change.

src/components/Board/Board.actions.js
```diff
@@ -89,6 +89,12 @@ function getActiveCommunicator(getState) {
   return activeCommunicator;
 }
 
+function replaceHistory(getState) {
```
tomivm: We should put a better name for this function. If it will replace the boardId by the activeBoardId we should reflect it in the name. the name could be `replaceHistoryWithActiveBoardId`

tomivm (changes requested):
Hey Gonza! It looks nice! I requested some changes. Also, I think we can rename the action 'toRootBoard' to a proper name in this PR. This action returns to the first board on navigationHistory and not to the communicator's rootBoard. We can probably choose a proper name.
I will test this and let u know if I find a problem :)
Thanks for your contribution. This code is not easy to refactor 
Also if you want my review in a PR please request my review to get notified

PR #1869, opened 2025-05-21
Fix fixed boards double render

### LINKED ISSUE #1868: Grid “blink” when `order` and `items` update separately in GridBase.js
## Description
When `GridBase` receives its props as separate updates—first a new `order`, then a new `items` array—it runs its internal [`sortGrid`](https://github.com/cboard-org/cboard/blob/1313d8d8cb058727f96c8ae7b38c11564ba78f74/src/components/FixedGrid/GridBase.js#L28) twice That results in two back-to-back renders and a visible “blink” as the grid reflows from the old state to the in-between state, then to the final state.

## Video

https://github.com/user-attachments/assets/82063a81-3643-40a9-9693-0ee65ad09ffe

RodriSanchez1:
close #1868

### DIFF
--- a/src/components/FixedGrid/Grid.js
+++ b/src/components/FixedGrid/Grid.js
@@ -212,7 +212,7 @@ function Grid(props) {
         setIsScroll(isScroll, totalRows);
       }
     },
-    [gridState.items, pages, setIsScroll, other.rows, isBigScrollBtns]
+    [items, pages, setIsScroll, other.rows, isBigScrollBtns]
   );
 
   return (
--- a/src/components/FixedGrid/Grid.js
+++ b/src/components/FixedGrid/Grid.js
@@ -31,15 +31,16 @@ function Grid(props) {
     gridState,
     ...other
   } = props;
+  const { items } = gridState;
 
   const itemsPerPage = other.rows * other.columns;
-  const [pages, setPages] = useState(chunks(gridState.items, itemsPerPage));
+  const [pages, setPages] = useState(chunks(items, itemsPerPage));
 
   useEffect(
     () => {
-      setPages(chunks(gridState.items, itemsPerPage));
+      setPages(chunks(items, itemsPerPage));
     },
-    [gridState.items, itemsPerPage]
+    [items, itemsPerPage]
   );
 
   const gridClassName = classNames(styles.grid, className);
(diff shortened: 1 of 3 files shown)


src/components/FixedGrid/GridBase.js
```diff
@@ -7,25 +7,33 @@ import Row from './Row/Row';
...
   const gridClassName = classNames(styles.root, className);
 
-  const grid = utils.sortGrid({ columns, rows, order, items });
+  const grid = useMemo(
+    () => {
+      console.log('REmemoized');
+      return utils.sortGrid({ columns, rows, order, items });
+    },
+    [columns, rows, order, items]
+  );
```
RodriSanchez1: This `useMemo`  is not working as expected because the `gridState` is defined on the `Board.component` as an inline component. It gets recreated when the BoardComponent is rendered. The memozation of this `gridState` is not avaible beacuse `Board` is a class componente and the hook `useMemo` is not avaliable´.  @tomivm @shayc  @gonojuarez what do you think?

src/components/FixedGrid/Grid.js
```diff
@@ -23,23 +23,23 @@ const focusPosition = {
...
     className,
-    items,
     style,
     setIsScroll,
     fixedRef,
     isBigScrollBtns,
     isNavigationButtonsOnTheSide,
+    gridState,
     ...other
   } = props;
```
tomivm, line 35: We can add ` const { items } = gridState;` and avoid lot of changes in this file

src/components/FixedGrid/GridBase.js
tomivm: Do you think this change is necessary in this PR? I think that we can treat it separately and merge this PR without inconvenience.
Let me know!

src/components/FixedGrid/Grid.js
RodriSanchez1: Yeah! Good idea

src/components/FixedGrid/GridBase.js
RodriSanchez1: We can leave this to other PR where we improve the grid performance. I will create a new issue to remember to work on this

RodriSanchez1: Anyway, I tested it without `useMemo` and noticed more re-renders than when we use `useMemo`, so I’ll leave it.

tomivm: Fine Rodri!! 🚀

tomivm (approved):
Awesome Job @RodriSanchez1 🚀

PR #1874, opened 2025-05-26
Enhance Cache-Control headers

### LINKED ISSUE #1482: After a release first Render of app don't work
`caught SyntaxError: Unexpected token '<'`

Way to repoduce it: 

1. Create a image build.
2. Run a docker container with this image
4. Open the app on localhost
5. Stop the container
6. Create a new image with some changes
7. Run the container with the newest image.
8. You will get the blank page with the `caught SyntaxError: Unexpected token '<'` on console

tomivm:
Refactor Cache-Control headers for HTML responses to improve caching behavior.
Close #1482

### DIFF
--- a/rootfs/etc/nginx/conf.d/default.conf
+++ b/rootfs/etc/nginx/conf.d/default.conf
@@ -14,8 +14,20 @@ server {
       log_not_found off;
     }
 
-    try_files $uri $uri/ /index.html;
-    
+    location ~* \.html$ {
+      add_header Cache-Control "no-cache, no-store, must-revalidate" always;
+    }
+
+    location = / {
+      add_header Cache-Control "no-cache, no-store, must-revalidate" always;
+      try_files $uri /index.html;
+    }
+
+    location / {
+      add_header Cache-Control "no-cache, no-store, must-revalidate" always;
+      try_files $uri $uri/ /index.html;
+    }
+
     set $auth_basic off;
     if (-f /auth/.htpasswd) {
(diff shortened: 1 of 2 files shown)


rootfs/etc/nginx/conf.d/default.conf
```diff
@@ -14,8 +14,20 @@ server {
...
-    try_files $uri $uri/ /index.html;
-    
+    location ~* \.html$ {
+      add_header Cache-Control "no-cache, no-store, must-revalidate" always;
+    }
+
+    location = / {
+      add_header Cache-Control "no-cache, no-store, must-revalidate" always;
+      try_files $uri /index.html;
+    }
```
RodriSanchez1, line 24: This block is repetead below

rootfs/etc/nginx/conf.d/default.conf
```diff
@@ -14,8 +14,20 @@ server {
...
+
+    location / {
+      add_header Cache-Control "no-cache, no-store, must-revalidate" always;
+      try_files $uri $uri/ /index.html;
+    }
```
RodriSanchez1: We can simplify this block just doing

``` suggestion
    location / {
        try_files $uri $uri/ /index.html;
    }
```
Why?
We don't wan to add the `add_header Cache-Control "no-cache, no-store, must-revalidate" always;` to every file on the project, just to the `.html`. 

The above `location ~* \.html$` will add the desired cached control. 

Just use  `location /` to catch all the request to the nginx proxy

PR #1876, opened 2025-05-27
Increase Service Worker Precache Size to Prevent Blank First Load

### LINKED ISSUE #1482: After a release first Render of app don't work

RodriSanchez1:
**Background**
On recent releases, users have encountered a blank page on their very first visit after an update. Investigation revealed that our service worker’s precache configuration was skipping the new `main.<hash>.js` bundle because it exceeded the default `maximumFileSizeToCacheInBytes` (≈ 4 MB). As a result:

1. The old service worker still served the cached `index.html` (pointing at the old bundle).
2. Meanwhile, the new worker installed in the background and cached the updated `index.html` (now referencing the new, larger `main.js`).
3. On reload, the service worker tried to serve the updated `index.html` but had no `main.js` in cache—leading to a blank shell until the browser fetched that script from the network (often requiring multiple refreshes).

**What Changed**

* Bumped `maximumFileSizeToCacheInBytes` in our `sw-precache` config from 4 MB to **8 MB**.

  ```js
  module.exports = {
    // …
    maximumFileSizeToCacheInBytes: 8 * 1024 * 1024, // now allows up to 8 MB bundles
    // …
  };
  ```
* This ensures that our primary JS bundle (`main.<hash>.js`) is included in the install‐time precache on every release.

**Why This Fixes the Blank Page**

* **Immediate availability**: With the larger threshold, the new `main.js` is fetched and stored during the service worker’s `install` event, alongside the updated `index.html`.
* **Consistent HTML↔JS mapping**: The precache now contains both the HTML and its referenced JS, so users no longer hit a mismatch on first load.
* **Full offline support**: On first load after an update—even if offline—the app shell and core scripts are already in cache, preventing the blank page scenario.

**Impact**

* Users will see the updated application immediately after each release, without needing repeated hard reloads.
* Slightly larger install-time download (up to 8 MB vs. 4 MB), but this is acceptable for faster, more reliable updates.

For reference [see](https://stackoverflow.com/a/76012729)

close #1482

### DIFF
--- a/sw-precache-config.js
+++ b/sw-precache-config.js
@@ -32,7 +32,7 @@ module.exports = {
     'build/static/**/*.*',
     ...boardImages
   ],
-  maximumFileSizeToCacheInBytes: 4194304,
+  maximumFileSizeToCacheInBytes: 8 * 1024 * 1024, // 8 MB
   runtimeCaching: [
     {
       urlPattern: /\/static\//,
--- a/sw-precache-config.js
+++ b/sw-precache-config.js
@@ -62,7 +62,7 @@ module.exports = {
       }
     }
   ],
-  navigateFallback: '/',
+  navigateFallback: '/index.html',
   dontCacheBustUrlsMatching: /\.\w{8}\./,
   dynamicUrlToDependencies: {
     '/': ['build/index.html']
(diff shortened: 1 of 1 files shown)


tomivm (approved):
Awesome research!

PR #1880, opened 2025-05-27
Fix Grid component to use pageItems in gridState

### LINKED ISSUE #1879: Board is duplicated on Fixed Grids with more than a page

tomivm:
Fix the Grid component to avoid duplication of first page.
Close #1879


RodriSanchez1 (approved):
Nice catch!

PR #1883, opened 2025-05-27
Optimise Board component

### LINKED ISSUE #1884: Avoid the use  of `translatedBoard` state on Board.container
This line of state inside our Board.container is causing a large number of rerenders of our boards.
We should avoid these to increase the performance of our app.
https://github.com/cboard-org/cboard/blob/109655b234cac88784a24d809d512b4659769a5e/src/components/Board/Board.container.js#L332

tomivm:
Remove unnecessary translatedBoard state and related logic from BoardContainer to improve performance and reduce re-renders.
Manage translations of labels when the Symbol component is rendered, providing a translation using labelKey. 
Please try to define if this way of translating labels is okay, if we can improve it, or find a better solution.

This PR drastically reduces the number of rerenders of our main `board` components.
Please let me know your thoughts, and thanks to @RodriSanchez1 for noting this performance issue.
Close #1884

### DIFF
--- a/src/components/Board/Board.container.js
+++ b/src/components/Board/Board.container.js
@@ -820,7 +820,11 @@ export class BoardContainer extends Component {
       if (tile.sound) {
         this.playAudio(tile.sound);
       } else {
-        const toSpeak = !hasAction ? tile.vocalization || tile.label : null;
+        const toSpeak =
+          (!hasAction && tile.vocalization) ||
+          tile.label ||
+          this.props.intl?.messages[tile.labelKey] ||
+          '';
         if (toSpeak) {
           speak(toSpeak);
         }
--- a/src/components/Board/Board.container.js
+++ b/src/components/Board/Board.container.js
@@ -193,7 +193,6 @@ export class BoardContainer extends Component {
     isSelecting: false,
     isLocked: true,
     tileEditorOpen: false,
-    translatedBoard: null,
     isGettingApiObjects: false,
     copyPublicBoard: false,
     blockedPrivateBoard: false,
@@ -285,9 +284,6 @@ export class BoardContainer extends Component {
     const goTo = id ? boardId : `board/${boardId}`;
     history.replace(goTo);
 
-    const translatedBoard = this.translateBoard(boardExists);
-    this.setState({ translatedBoard });
-
     //set board type
     this.setState({ isFixedBoard: !!boardExists.isFixed });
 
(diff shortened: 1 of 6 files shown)


src/helpers.js
```diff
@@ -48,3 +48,11 @@ export const convertImageUrlToCatchable = imageUrl => {
   if (isCboardProductionBlobContainer) return cboardBlobUsingCDN(imageUrl);
   return null;
 };
+
+export const resolveLabel = ({ label, labelKey }, intl) => {
```
shayc: The name `resolveLabel` is too generic, especially for a helper util that will be imported in many files.
I suggest using something more specific, like `resolveTileLabel`.

src/components/Board/EditToolbar/EditToolbar.component.js
```diff
@@ -135,7 +137,7 @@ function EditToolbar({
           })}
           onClick={onBoardTitleClick}
         >
-          {board.name}
+          {board.name || intl.formatMessage({ id: board.nameKey })}
```
shayc: Maybe extract this logic to a helper like `resolveLabel`, perhaps `resolveBoardName`

src/components/Board/Board.container.js
```diff
@@ -193,7 +194,6 @@ export class BoardContainer extends Component {
     isSelecting: false,
     isLocked: true,
     tileEditorOpen: false,
-    translatedBoard: null,
```
shayc, line 196: I love seeing a bunch of code getting deleted 😍

shayc (commented):
This is awesome!

src/helpers.js
RodriSanchez1: I agree!

src/components/Board/EditToolbar/EditToolbar.component.js
RodriSanchez1: Like that!

src/helpers.js
tomivm: True!!

src/components/Board/Board.container.js
tomivm: Hehe was a pleasure to press 'delete'

tomivm:
Thanks for your review @shayc !!!

src/components/Board/EditToolbar/EditToolbar.component.js
tomivm: :)

RodriSanchez1 (approved):
Grat jobbbbbbbbb!

Issue #1878, opened 2025-05-27
Fix FixedGrid `chunks` function to separate the tiles depending on order


tomivm:
Now the function is separating the latest tiles of the `items` using the `itemsPerPages`, but we need to separate by the `order`.
example: If we remove the last row, we should separate the elements that aren't in the new order.
https://github.com/cboard-org/cboard/blob/b8a35540099224609fe8109ec02052573b5005ab/src/components/FixedGrid/Grid.js#L8

tomivm:
Idealy We should implement. this function
```js
export function chunks({ tileItems, order }:{
  tileItems: Array<TileItem>;
  order: GridOrder;
}) {
  const firstPageItemsInOrder = order.map(row =>
    row.map(id => tileItems.find(item => item.id === id))
  );

  const firstPage = lodash.flatten(firstPageItemsInOrder);

  const size = firstPage.length;

  const restOfItems = tileItems.filter(
    item => !firstPage.find(firstItem => firstItem?.id === item.id)
  );
  const restOfPages = lodash.chunk(restOfItems, size);

  return [firstPage, ...restOfPages];
}
```
But this causes differences between how the users see his boards now and with this function.
For that I refactorized it to have a better chunk in pages but preserve the functionability of fill the board with remaining tiles.

PR #1891, opened 2025-05-29
Quick fix for remove rows and columns function on fixed boards

### LINKED ISSUE #1109: Fixed board - Add or remove row or colum dont change grid order prop  correctly
When you add or remove a row or colum  on the fixed board, the grid order prop is not changed correctly. In the image, you can visualice the tiles position arent the same than the order displayed in the redux state.
image

tomivm:
Introduce a functions to chunk fixed boards efficiently while maintaining compatibility with deprecated chunking methods that cause 'order' issues when users removes columns or rows.
Now when we remove rows or columns the board conserve it's structure but remain filling empty spaces with tiles.
This is a quick fix until solve #1878
Close #1109

### DIFF
--- a/src/components/FixedGrid/utils.ts
+++ b/src/components/FixedGrid/utils.ts
@@ -138,33 +138,36 @@ interface TileItem {
   linkedBoard?: boolean;
 }
 
-export function deprecatedChunks(tileItems: Array<TileItem>, size: number) {
-  const newArray = [...tileItems];
-  const results = [];
-
-  while (newArray.length) {
-    results.push(newArray.splice(0, size));
-  }
-
-  return results;
-}
-
-export function chunks({ tileItems, order }:{
+export function compatibleDeprecatedChunks({ tileItems, order }:{
   tileItems: Array<TileItem>;
   order: GridOrder;
-}) {
+}):TileItem[][] {
   const firstPageItemsInOrder = order.map(row =>
-    row.map(id => tileItems.find(item => item.id === id))
+    row.map(id => tileItems.find(item => item.id === id)||null)
   );
 
   const firstPage = lodash.flatten(firstPageItemsInOrder);
+  const unnorderedTiles = tileItems.filter(
+    item => !firstPage.find(tile => tile?.id === item.id)
+  );
+  const fillEmptyPositionsWithUnnorderedTilesForOldBoards = ({firstPageItemsInOrder,unnorderedTiles}:{firstPageItemsInOrder:(TileItem|null)[][],unnorderedTiles:Array<TileItem> }) => {
+    let index = 0;
+    const deprecatedFirstPageItemsInOrder = firstPageItemsInOrder.map(row=>
+    row.map(item => {
+      if (item) return item;
+      index++;
+      return unnorderedTiles[index - 1]|| null;
+    }));
+    const deprecatedFirstPage = lodash.flatten(deprecatedFirstPageItemsInOrder).filter(item => item !== null);//
+    const restOfTiles = unnorderedTiles.slice(index);
+
+    return {deprecatedFirstPage,restOfTiles};
+  }
+  
+  const {deprecatedFirstPage ,restOfTiles} = fillEmptyPositionsWithUnnorderedTilesForOldBoards({firstPageItemsInOrder,unnorderedTiles});
 
   const size = firstPage.length;
-
-  const restOfItems = tileItems.filter(
-    item => !firstPage.find(firstItem => firstItem?.id === item.id)
-  );
-  const restOfPages = lodash.chunk(restOfItems, size);
-
-  return [firstPage, ...restOfPages];
+  const restOfPages = lodash.chunk(restOfTiles, size);
+  
+  return [deprecatedFirstPage, ...restOfPages];
 }
--- a/src/components/FixedGrid/utils.ts
+++ b/src/components/FixedGrid/utils.ts
@@ -183,8 +183,8 @@ export function getTilesListForNewOrder({
   order: GridOrder;
 }): TileItem[] {
   const newPages = getDeprecatedOrderedPages({
-    tileItems,
     order,
+    tileItems,
   });
   const tilesListForNewOrder = newPages.flat();
   return tilesListForNewOrder;
--- a/src/components/FixedGrid/utils.ts
+++ b/src/components/FixedGrid/utils.ts
@@ -126,32 +126,48 @@ export function removeOrderItems(ids: string, order: GridOrder): GridOrder {
   return order.map(row => row.map(id => (id && ids.includes(id) ? null : id)));
 }
 
-function getDeprecatedOrderedPages({ tileItems, order }:{
+function getDeprecatedOrderedPages({
+  tileItems,
+  order,
+}: {
   tileItems: Array<TileItem>;
   order: GridOrder;
-}):TileItem[][] {
-  const firstPageItemsInOrder = order.map(row =>
-    row.map(id => tileItems.find(item => item.id === id)||null)
+}): TileItem[][] {
+  const firstPageItemsInOrder = order.map((row) =>
+    row.map((id) => tileItems.find((item) => item.id === id) || null),
   );
 
   const firstPageIds = order.flat();
   const orderedIds = new Set(order.flat());
-  const unorderedTiles = tileItems.filter(item => !orderedIds.has(item.id));
-  const fillEmptyPositionsWithUnorderedTilesForOldBoards = ({firstPageItemsInOrder,unorderedTiles}:{firstPageItemsInOrder:(TileItem|null)[][],unorderedTiles:Array<TileItem> }) => {
+  const unorderedTiles = tileItems.filter((item) => !orderedIds.has(item.id));
+  const fillEmptyPositionsWithUnorderedTilesForOldBoards = ({
+    firstPageItemsInOrder,
+    unorderedTiles,
+  }: {
+    firstPageItemsInOrder: (TileItem | null)[][];
+    unorderedTiles: Array<TileItem>;
+  }) => {
     let index = 0;
-    const deprecatedFirstPageItemsInOrder = firstPageItemsInOrder.map(row=>
-    row.map(item => {
+    const deprecatedFirstPageItemsInOrder = firstPageItemsInOrder.map((row) =>
+      row.map((item) => {
       if (item) return item;
       index++;
-      return unorderedTiles[index - 1]|| null;
-    }));
-    const deprecatedFirstPage = deprecatedFirstPageItemsInOrder.flat().filter(item => item !== null);
+        return unorderedTiles[index - 1] || null;
+      }),
+    );
+    const deprecatedFirstPage = deprecatedFirstPageItemsInOrder
+      .flat()
+      .filter((item) => item !== null);
     const restOfTiles = unorderedTiles.slice(index);
 
-    return {deprecatedFirstPage,restOfTiles};
-  }
+    return { deprecatedFirstPage, restOfTiles };
+  };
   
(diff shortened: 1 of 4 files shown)


src/components/FixedGrid/utils.ts
```diff
@@ -123,3 +124,50 @@ export function getNewOrder({ columns,
...
+  label?: string;
+  labelKey?: string;
+  vocalization?: string;
+  image: string;
+  loadBoard?: string;
+  sound?: string;
+  type?: string;
+  backgroundColor: string;
+  linkedBoard?: boolean;
+}
```
RodriSanchez1: Relocate this interface to a new type file. As this it gonna be used across all the app across the progresive migration to TS

src/components/FixedGrid/utils.ts
```diff
@@ -123,3 +124,50 @@ export function getNewOrder({ columns,
...
+
+export function compatibleDeprecatedChunks({ tileItems, order }:{
```
RodriSanchez1: We should rename this function to be more descriptive as it return the ordered pages

src/components/FixedGrid/utils.ts
```diff
@@ -123,3 +124,50 @@ export function getNewOrder({ columns,
...
+  tileItems: Array<TileItem>;
+  order: GridOrder;
+}):TileItem[][] {
+  const firstPageItemsInOrder = order.map(row =>
+    row.map(id => tileItems.find(item => item.id === id)||null)
+  );
```
RodriSanchez1: Consider use a hash map to improve the performance of this operations and make it faster. Take a look on [js map](https://www.geeksforgeeks.org/javascript-map/) and [Hash Maps to Improve Time Complexity](https://javascript.plainenglish.io/hash-maps-to-improve-time-complexity-bc7ca692fcc) 

```js
const tileMap = new Map(tileItems.map(item => [item.id, item]));
const firstPageItemsInOrder = order.map(row =>
    row.map(id => tileMap.get(id) || null)
    );

```

src/components/FixedGrid/utils.ts
```diff
@@ -123,3 +124,50 @@ export function getNewOrder({ columns,
...
+
+  const firstPage = lodash.flatten(firstPageItemsInOrder);
+  const unnorderedTiles = tileItems.filter(
```
RodriSanchez1: Fix typo - `unorderedTiles`

src/components/FixedGrid/utils.ts
```diff
@@ -123,3 +124,50 @@ export function getNewOrder({ columns,
...
+    item => !firstPage.find(tile => tile?.id === item.id)
+  );
+  const fillEmptyPositionsWithUnnorderedTilesForOldBoards = ({firstPageItemsInOrder,unnorderedTiles}:{firstPageItemsInOrder:(TileItem|null)[][],unnorderedTiles:Array<TileItem> }) => {
```
RodriSanchez1: Fix typo `fillEmptyPositionsWithUnorderedTilesForOldBoards`

src/components/FixedGrid/utils.ts
```diff
@@ -123,3 +124,50 @@ export function getNewOrder({ columns,
...
+
+export function compatibleDeprecatedChunks({ tileItems, order }:{
+  tileItems: Array<TileItem>;
+  order: GridOrder;
+}):TileItem[][] {
+  const firstPageItemsInOrder = order.map(row =>
+    row.map(id => tileItems.find(item => item.id === id)||null)
+  );
+
+  const firstPage = lodash.flatten(firstPageItemsInOrder);
```
RodriSanchez1: Why dont you use the  [Array.prototype.flat()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat)?

src/components/FixedGrid/utils.ts
```diff
@@ -123,3 +124,50 @@ export function getNewOrder({ columns,
...
+  const unnorderedTiles = tileItems.filter(
+    item => !firstPage.find(tile => tile?.id === item.id)
+  );
```
RodriSanchez1: You can use the [Set ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) to save the ordered tile Ids and use it to find the unordered tiles faster 

```js
const orderedIds = new Set(order.flat()); // or you can do the new Set(firstPage.flaten())
const unorderedTiles = tileItems.filter(item => !orderedIds.has(item.id));
```

RodriSanchez1 (changes requested):
Its a good idea to start refactoring the Fixed Grids. The `Grid` and the `Grid Base` can be merged in one file and we gonna have just one component that mange the grid render. 

What happens with the GridBase component if we merge this PR? I will recompute the Grid again within the sortGrid?

src/components/FixedGrid/utils.ts
```diff
@@ -123,3 +124,50 @@ export function getNewOrder({ columns,
...
+  const fillEmptyPositionsWithUnnorderedTilesForOldBoards = ({firstPageItemsInOrder,unnorderedTiles}:{firstPageItemsInOrder:(TileItem|null)[][],unnorderedTiles:Array<TileItem> }) => {
+    let index = 0;
+    const deprecatedFirstPageItemsInOrder = firstPageItemsInOrder.map(row=>
+    row.map(item => {
+      if (item) return item;
+      index++;
+      return unnorderedTiles[index - 1]|| null;
+    }));
+    const deprecatedFirstPage = lodash.flatten(deprecatedFirstPageItemsInOrder).filter(item => item !== null);//
```
RodriSanchez1: Consider using the Array.prototype.flat()

RodriSanchez1:
Plese fix the [Codacy Static Code Analysis](https://github.com/cboard-org/cboard/pull/1891/checks?check_run_id=43150601881)

src/components/FixedGrid/utils.ts
tomivm: Because is experimental

src/components/FixedGrid/utils.ts
tomivm: Looks awesome but as is not a critical task I will create a ticket to give me time first to learn how to use it

src/components/FixedGrid/utils.ts
tomivm: Can you recommend a name please

tomivm:
Hey @RodriSanchez1 
Thats is the plan :) Please for that. Can you please put more focus on the functionality and to review if we don't have differences between the new and the past behavior.
Next week I will focus on fix the typos and review your performance increase suggestion
My prettier is failing for this project in TS. Could you format the file to avoid the failing statics tests? 🔴

src/components/FixedGrid/utils.ts
RodriSanchez1: Not really [see compatibility](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat#browser_compatibility)

RodriSanchez1:
> Its a good idea to start refactoring the Fixed Grids. The `Grid` and the `Grid Base` can be merged in one file and we gonna have just one component that mange the grid render.
> 
> What happens with the GridBase component if we merge this PR? I will recompute the Grid again within the sortGrid?

yeah, could you respond that?

The prettier is failing for me also. Is something on the TS config version and the file. We should fix this in the project

tomivm:
Sorry I give you a more detailed explanation 💥 
Yes that is the plan! 
We gonna implement this in small steps focusing in Continues Integration.
The main Idea of this PR is to fix users intents that remove columns or rows, to not destroy his boards.
This is archived by refactoring a wrong chunk function that is cutting our pages only using the size of the board but ignoring the order. Also if we use only the order to separate pages, the tiles with not order (that is a bug that we have now in our board) are moved to the second page but it causes changes in how users with bugged boards are seeing his tiles.
For that we should threat this as a "pure function" to start having the same behaviour and continuing passing for the sortGrid to prevent bug in edge cases until we can plan a migration to a better gridSorting that probably would be do it by "merging the Grid and the Grid Base" as you said

src/components/FixedGrid/utils.ts
RodriSanchez1: Something like `getDeprecatedOrderedPages` ?

RodriSanchez1:
Okay, thanks for the explanation. It makes sense. One think that I don't like is to repeat the sorting on the chunk function and after that in the `sortGrid` , but if the idea is to refactor the `GridBase` soon it can be okay.

Regarding to the PR, I made a basic test: Render a fixed board using the old chunks and the new one. This should render the same pages as it supports the old  deprecated behaviour . 

As you can see in the following image the rendered pages are not the same. 
image

The attached image is not accurate to the following reproduction steps, but you will get the same:

1- Clean browser cache
2- Go to PicseePal board
3- Remove a row
4- See the rendered pages
5- Go to the code and make the changes to use the old chunk function
6- Compare the rendered pages with the previous ones

The exported are different also, it think it use the old behaviour to generate the exported pdf

Sorry for the late response. Is a complex functionallity and a extensive test should be done before merge it to master. 

I like the idea to refactor this,

src/components/FixedGrid/utils.ts
tomivm: Awesome! I saw and old documentation

src/components/FixedGrid/utils.ts
tomivm: Perfect!

src/components/FixedGrid/utils.ts
tomivm: Nice suggestion! thanks

tomivm:
Hey Rodri! Awesome feedback! That is the valuable feedback that I was looking for. 
As you said this is a complex functionality and for that your feedback is important. Also for that the time that we take to review PRs directly impact in the continuos integration that our team is trying to archive! 
If the review is delayed the complex of the issue cause developer have to invert time to remember the code, prepare the enviroment causing it's first read of the review have less context and more tasks to do.
It's okey to take time to make reviews with dedication but please the next time that **a partner is awaiting something from you** don't spent your time in another issue with low priority until you finish supporting the team work.
Thanks for dedicate the time to this! you found a case that I never thought for this. 
**Users make a change in a new version of the app and have other devices with old versions**
Also it represent your worry of "I don't like is to repeat the sorting on the chunk function and after that in the sortGrid"
This makes me think that if we want to improve sorting and we can dedicate a full PR to intent to move the sorting logic up
For that to solve this and also solve the problem that only occurs when a column or a row is removed I move the logic up to the board container and the sort function is only called inside `handleAddRemoveRow` and `handleAddRemoveColumn`
This fix the order of the tileList ensuring old versions continue showing the exact same board than a new version and ensure that this changes only affects the functionality that have the issue and avoid unnecessary changes to our Grid component that caused the unnecessary double sorting when props `order` and `tiles` changes.
As you said this functionality needs an extensive test. Please review it as soon as possible to be integrated before the testing process pre-deploy planed for this week! Also i contacted a Speech therapist that work with Cboard for long time to test all the improvements that we make the past week.
Sadly we don't know when we can dedicate a full week to Cboard again to improve the sortGrid process and this change drastically improve the break of user's boards during the use of a feature that is in master!!

Again, Thanks for your feedback and support that allow Cboard to continue growing and improving day by day 🥳 
Let me know your thoughts please!

RodriSanchez1 (approved):
Lot of things happened.  I thinking what side effectes we can get if the tiles list changes. For now a I was not able to find anything that can breaks. 

This is a good provisional solution, well done 🤘! 

The referenced issue is not longer apropiate for this PR :(

tomivm:
Yes I find an old issue about this. The issue of the chunking issue was only mentioned
Happy of launching this together with all the improvements that we make in our next version!

PR #1898, opened 2025-05-30
Refactor image source handling in Symbol 

### LINKED ISSUE #1892: Tile images blink when switch between folders
If this is the first time opening a board, it's normal for the images on the tiles to be empty until they load. However, on subsequent visits, the images should be available quickly and without blinking.

RodriSanchez1:
A new function to format the image src was added. This were used on the `src` state initialization. On this way, the Symbol has a value for the src on firs render. This avoid the image blink between board navegations

close #1892


tomivm (approved):
Good eye! you find a really important detail!

PR #1899, opened 2025-05-30
Fix memory leak and improve image fetching in Symbol component

### LINKED ISSUE #1900: Symbols images URLs are not being cleaned on component unmount
This can generate memory leaks

### LINKED ISSUE #1901: Refactor Symbol component
Should: 

- Handle errors on indexed db call and image retrival
- Improve the useEffect block to use early returns
- Add a cancelled state that should avoid state changes when the component is unmounted preventing warnings

RodriSanchez1:
See first the  #1898 

Address memory leak by revoking object URLs in the cleanup function and streamline image fetching from the indexed DB. Remove unnecessary dependencies from the useEffect hook and improve code legibility using early returns.

close #1901
close #1900

### DIFF
--- a/src/components/Board/Symbol/Symbol.js
+++ b/src/components/Board/Symbol/Symbol.js
@@ -91,7 +91,6 @@ function Symbol(props) {
       }
       getSrc();
 
-      // Cleanup function
       return () => {
         cancelled = true;
         if (objectUrlRef.current) {
--- a/src/components/Board/Symbol/Symbol.js
+++ b/src/components/Board/Symbol/Symbol.js
@@ -72,7 +72,7 @@ function Symbol(props) {
         }
       };
     },
-    [keyPath, setSrc, props.image]
+    [keyPath, props.image]
   );
 
   const symbolClassName = classNames('Symbol', className);
--- a/src/components/Board/Symbol/Symbol.js
+++ b/src/components/Board/Symbol/Symbol.js
@@ -43,6 +43,8 @@ function Symbol(props) {
   } = props;
 
   const [src, setSrc] = useState(image ? formatSrc(image) : '');
+  const objectUrlRef = useRef(null);
+
   const fetchArasaacImage = useCallback(async id => {
     if (!id) return null;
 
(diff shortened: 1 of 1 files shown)


src/components/Board/Symbol/Symbol.js
```diff
@@ -43,27 +43,62 @@ function Symbol(props) {
   } = props;
 
   const [src, setSrc] = useState(image ? formatSrc(image) : '');
+  const objectUrlRef = useRef(null);
+
+  const fetchArasaacImage = useCallback(async id => {
```
tomivm: We can name it with something similar to`fetchArasaacImagefromIndexedDB` or `fetchLocalArasaacImage`

src/components/Board/Symbol/Symbol.js
```diff
@@ -43,27 +43,62 @@ function Symbol(props) {
...
-          const blob = new Blob([image.data], { type: image.type });
-          setSrc(URL.createObjectURL(blob));
-        } else if (props.image) {
-          setSrc(formatSrc(props.image));
+          setSrc(formatSrc(image));
+          return;
         }
-      }
 
+        setSrc('');
```
tomivm, line 90: this is necessary?

src/components/Board/Symbol/Symbol.js
```diff
@@ -43,27 +43,62 @@ function Symbol(props) {
...
+      }
       getSrc();
+
+      // Cleanup function
+      return () => {
+        cancelled = true;
```
tomivm, line 95: Can you explain me why the cancelled is necessary?

src/components/Board/Symbol/Symbol.js
```diff
@@ -43,27 +43,62 @@ function Symbol(props) {
...
+          setSrc(formatSrc(image));
+          return;
         }
-      }
 
+        setSrc('');
+      }
       getSrc();
+
+      // Cleanup function
```
tomivm: remove comment please :)

tomivm (approved):
Looks good! We can think in remove the image from the useEffect dependencies. But maybe is better to do it in a next step to avoid edge cases for now

src/components/Board/Symbol/Symbol.js
RodriSanchez1: Is a fallback, if not an old image can be showed to the user and it can be confusing

src/components/Board/Symbol/Symbol.js
RodriSanchez1: This `cancelled` variable is neccesary to avoid changes in the component state if it was unmounted while an async function was performed.

RodriSanchez1:
Thanks for the review! How do you think that we can remove the image from the dependecies? We need to manage the case when a tile have `image` and `local Arasaac image`

src/components/Board/Symbol/Symbol.js
tomivm: Fine thanks!

tomivm:
Maybe we can consider show the local first if it is available. 
I think we can discuss it in another issue.
Awesome job!

tomivm:
Sorry @RodriSanchez1 I miss that this branch is not pointing to the master branch. It's okey?

RodriSanchez1:
It was a mistake! I just created a new PR to the master branch

PR #1906, opened 2025-06-03
Fix picseepal board tiles grid order

### LINKED ISSUE #1905: Some tiles on our default board have an undefined position in the grid order
We need to indicate the correct position for tiles in our Core Pic See Pal boards.
See the next commit to understand how.
https://github.com/cboard-org/cboard/commit/eaf784fefaf27bdd46f28e8cb947d92b0dbb156c

RodriSanchez1:
close #1905


tomivm (approved):
Nice!!

Issue #1896, opened 2025-05-30
Add ts files to Prettier formatting


RodriSanchez1:

zkarmi:
Hello @RodriSanchez1 , I'm happy to take this. This would be my first open source contribution, so thank you for the opportunity, if available.

I've locally updated `package.json` to include TS files in the `lint-staged` pre-commit. When running lint-staged after this update, it fails:

![Image](https://github.com/user-attachments/assets/89e3b451-e3bb-4276-bb10-7b6d316fca67)

The version of Prettier being used (1.15.3) does not support optional chaining, which became available in Prettier v1.19 (which added TS 3.7 support -- https://prettier.io/blog/2019/11/09/1.19.0.html). I tested with bumping to Prettier @2.0.0 and the commit works thereafter. The latest Prettier is v3.5.3, which we can also upgrade to. It does require Node 14.18.0 or later, if that's of concern.

Lastly, we can remove `git add` from the the lint-staged pre-commit because it's unnecessary as of v10 and triggers a warning (seen in the above screenshot).

RodriSanchez1:
Hi @zkarmi ! Welcome to the Cboard project! 

That's awesome! Thanks for that explanation. I think the better is to bump to the latest Prettier version. Please double check if there is some breaking changes between the actual and the latest versions. We dont want unexpected behaviour hehe. 

Remove the `git add` if is not necessary!

I will assign the issue to you. Thanks for your contribution

zkarmi:
@RodriSanchez1 Can you help me understand what breaking issues we're looking for here? My understanding is that the pre-commit will only lint staged files, so the remainder of the repo should stay unaffected by this change until individual files are updated.

I believe this Prettier config should handle keeping a number of the Prettier options as their original defaults from 1.15.3, if that makes things easier in case of PRs with large numbers of files. However it won't address any newer options that didn't exist then, such as `quoteProps`:
```json
{
  "printWidth": 80,
  "singleQuote": false,
  "tabWidth": 2,
  "useTabs": false,
  "semi": true,
  "trailingComma": "none",
  "bracketSpacing": true,
  "arrowParens": "avoid",
  "endOfLine": "auto",
  "jsxSingleQuote": false,
  "proseWrap": "preserve",
  "jsxBracketSameLine": false
}
```

[...]

PR #1923, opened 2025-06-17
Add TS files to Prettier formatting - Issue #1896

zkarmi:
Per discussion in #1896 this update enables Prettier on `.ts` and `.tsx` files.
The previous version of Prettier was quite outdated so this update bumps that version, and removes the redundant `git add` from the `lint-staged` hook that hasn't been needed since v10.
As I [mentioned in the issue](https://github.com/cboard-org/cboard/issues/1896#issuecomment-2969034466), I can also introduce some rules in `.prettierrc` to approximate previous behavior and reduce the number of changes when committing updates moving forward, if desired.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -10,5 +10,5 @@
   "endOfLine": "auto",
   "jsxSingleQuote": false,
   "proseWrap": "preserve",
-  "jsxBracketSameLine": false
+  "bracketSameLine": false
 }
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,4 +1,14 @@
 {
   "printWidth": 80,
-  "singleQuote": true
+  "singleQuote": true,
+  "tabWidth": 2,
+  "useTabs": false,
+  "semi": true,
+  "trailingComma": "none",
+  "bracketSpacing": true,
+  "arrowParens": "avoid",
+  "endOfLine": "auto",
+  "jsxSingleQuote": false,
+  "proseWrap": "preserve",
+  "jsxBracketSameLine": false
 }
--- a/package.json
+++ b/package.json
@@ -104,7 +104,7 @@
     "jest-mock-axios": "^3.2.0",
     "lint-staged": "^10.0.3",
     "node-polyfill-webpack-plugin": "^2.0.1",
-    "prettier": "1.15.3",
+    "prettier": "^3.5.3",
     "react-scripts": "5.0.1",
     "react-test-renderer": "^17.0.2",
     "redux-mock-store": "^1.5.4",
@@ -118,9 +118,8 @@
     }
   },
   "lint-staged": {
-    "*.{js,json,css}": [
-      "prettier --write",
-      "git add"
+    "*.{js,ts,tsx,json,css}": [
+      "prettier --write"
     ]
   },
   "jest": {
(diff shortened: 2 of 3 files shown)


RodriSanchez1:
Hi @zkarmi! This is awesome—thank you! I just saw the issue comment you left. I agree, we should implement the suggested changes in the .prettierrc file. Otherwise, too many formatting changes are introduced whenever an old file is saved.

I'll check with @tomivm about how we want to handle reformatting existing files. My initial thought is to create a separate commit for formatting changes in upcoming PRs. Let me know if you have any suggestion

zkarmi:
@RodriSanchez1 Glad I can help. I'm adding a commit for the expanded `.prettierrc`. I'm swapping out `"jsxBracketSameLine": false` for `"bracketSameLine": false`, as the former is deprecated.

Are you suggesting to first commit to apply formatting to the file, then make another commit with actual changes? That could make things cleaner when reviewing PRs, and you wouldn't have to turn off `lint-staged`. It _would_ mean you have to introduce that into the PR workflow, though.

RodriSanchez1:
> Are you suggesting to first commit to apply formatting to the file, then make another commit with the actual changes? That could make things cleaner when reviewing PRs, and you wouldn't have to turn off lint-staged.

Yes, that's exactly what I had in mind.

> It would mean you have to introduce that into the PR workflow, though.

That's a good point. I'll think about whether there's a better way to handle this. But either way, we'll need to push the new formatting at some point., so it might be cleaner to address it proactively.

RodriSanchez1:
Hey @zkarmi!

Today we had the Cboard weekly dev meeting and talked about the formatting modifications. We concluded that we are going to implement them in one PR, probably this one, where the whole project will be reformatted.

Next steps:

I'm going to do further research and check if there's any other property we can add to the `.prettierrc` file. I know you already did  that, but if there's anything you think I should look into or any documentation I should check, let me know.

We also need to double-check what will happen with the open PRs. Are we going to face any problems or conflicts if we apply the new formatting? If that's the case, the best would be to close as many PRs as possible before merging this one.

zkarmi:
Thanks for the update, @RodriSanchez1 !
I've not done such a Prettier migration myself previously, but from some research online the general consensus I've seen has been to do it all in one big PR, as you said. And yes, it seems like suggestions are to wrap up PRs and not create new ones while updating, to avoid merge conflicts.
While doing some reading I did come across this article where the author's team used some scripts to make the changes, allowing the `git blame` to fall on a tool rather than a specific dev. Maybe it will be helpful to the team, I wanted to share because I haven't had time to dive deep on it myself: https://flexport.engineering/upgrading-prettier-on-a-large-codebase-28d56c4de49e.

zkarmi:
Just to add here -- I would recommend reviewing which props were added after 1.15. The docs do show which version an option first appeared in: https://prettier.io/docs/options

RodriSanchez1:
That's awesome! Thanks for sharing. The team was right hehe. I will be checking out this week!

magush27:
I've tested this PR and can confirm the technical implementation is working correctly:

**What I tested:**
1. Prettier Version Update
   - Verified Prettier upgraded from v1.15.3 to v3.5.3
   - No more warnings about unknown options (e.g. bracketSameLine)
2. TypeScript File Formatting
   - Tested formatting on .ts files
   - IDE auto-format on save works correctly with the new configuration
   - Manual formatting via `yarn prettier --write` works without warnings
3. Pre-commit Hook (lint-staged)
   - Confirmed lint-staged now processes `*.{js,ts,tsx,json,css}` files
   - Pre-commit hook successfully formats TypeScript files before commit
   - Verified the redundant `git add` was removed (automatic since lint-staged v10)
4. Test Suite
   - Ran existing tests with no regressions related to Prettier changes

**Result:**
The technical implementation works as expected. The Prettier upgrade and TypeScript formatting are functioning correctly.

I understand from the discussion that the we are planning to finalize the .prettierrc configuration and coordinate the full codebase reformatting before merging. This test confirms the tooling changes are ready.

shayc:
#2316 + #2317 closes this and updates `husky` + `lint-staged`

shayc:
I wanted to take a moment and thank you @zkarmi
we didn't merge this pr at the end but your contribution is really genuinely appreciated.

PR #1930, opened 2025-06-17
AuthScreen redirects to root board on close

### LINKED ISSUE #1902: Fix Close action on Login Screen
Original: 
_I found a bug related to the close action at the login screen.
Supposed to return to the root board, not turn to the previous screen_

EDIT:  To reproduce this, acces to the `https://app.cboard.io/login-signup` and press the close (X) button. An backward action is dispatched insted to go to the root board or any board.

zkarmi:
This fixes #1902 so that when the `AuthScreen` component (`/login-signup`) is closed via the close (X) button, the user is taken to the root board.
To do this I've replaced the WelcomeScreen onClose prop with `history.replace('/board/root')` instead of `history.goBack`.

### DIFF
--- a/src/components/AuthScreen/AuthScreen.component.js
+++ b/src/components/AuthScreen/AuthScreen.component.js
@@ -6,7 +6,7 @@ import './AuthScreen.css';
 
 class AuthScreen extends Component {
   static propTypes = {
-    intl: intlShape.isRequired
+    intl: intlShape.isRequired,
   };
 
   render() {
@@ -17,7 +17,7 @@ class AuthScreen extends Component {
         <WelcomeScreen
           heading={intl.formatMessage(messages.heading)}
           text={intl.formatMessage(messages.text)}
-          onClose={history.goBack}
+          onClose={() => history.replace('/board/root')}
         />
       </div>
     );
--- a/src/components/AuthScreen/AuthScreen.component.js
+++ b/src/components/AuthScreen/AuthScreen.component.js
@@ -6,7 +6,7 @@ import './AuthScreen.css';
 
 class AuthScreen extends Component {
   static propTypes = {
-    intl: intlShape.isRequired,
+    intl: intlShape.isRequired
   };
 
   render() {
@@ -17,7 +17,9 @@ class AuthScreen extends Component {
         <WelcomeScreen
           heading={intl.formatMessage(messages.heading)}
           text={intl.formatMessage(messages.text)}
-          onClose={() => history.replace('/board/root')}
+          onClose={() => {
+            history.action === 'PUSH' ? history.goBack() : history.push('/');
+          }}
         />
       </div>
     );


src/components/AuthScreen/AuthScreen.component.js
```diff
@@ -6,7 +6,7 @@ import './AuthScreen.css';
 
 class AuthScreen extends Component {
   static propTypes = {
-    intl: intlShape.isRequired
+    intl: intlShape.isRequired,
```
RodriSanchez1: Please avoid to commit this change as is this change is not neccesary for this PR ;)

src/components/AuthScreen/AuthScreen.component.js
```diff
@@ -17,7 +17,7 @@ class AuthScreen extends Component {
         <WelcomeScreen
           heading={intl.formatMessage(messages.heading)}
           text={intl.formatMessage(messages.text)}
-          onClose={history.goBack}
+          onClose={() => history.replace('/board/root')}
```
RodriSanchez1: We should add a conditional check:

If the history stack contains a previous URL, use goBack to return the user to that URL (for example, when the login/signup flow is opened from the settings page).

If there is no previous URL in the history stack, redirect the user to the home page `history.replace('/');`
Please also make sure the navigation is properly performed and that the URL actually changes as expected.

RodriSanchez1 (changes requested):
Hi @zkarmi Welcome to the Cboard project! Thanks for your contribution

src/components/AuthScreen/AuthScreen.component.js
zkarmi: I believe this happened because Prettier got updated somehow in the branch I was working on. I'll roll this back.

src/components/AuthScreen/AuthScreen.component.js
zkarmi: @RodriSanchez1 Thank you for the feedback! This was my initial thought too in terms of user experience, but I understood the issue description as "_always_ go to the root board when the AuthScreen is closed":
> I found a bug related to the close action at the login screen.
> Supposed to return to the root board, not turn to the previous screen

The `history` object gives us a length but no visibility into the previous URL. Given that, we don't know _what_ the user's previous page was. It could have been the Settings page, but it could also have been google.com or even just the browser's New Tab screen. I want to confirm if it's OK to send the user back to the previous page as long as `history.length > 1`, even if we don't know what URL is at the top of the stack.

RodriSanchez1: Thanks for your feedback! I apologize, I hadn’t read the issue in detail earlier.

I’ve been investigating the `history` library to provide an accurate solution. As you pointed out, relying solely on `history.length` might not always result in reliable navigation behavior.

While reviewing the library, I noticed that `history` exposes an `action` property, which can be `POP`, `REPLACE`, or `PUSH`. (See the [action documentation](https://github.com/remix-run/history/blob/dev/docs/api-reference.md#action) for more details. Note: This is for version 5; we’re currently on version 4, but the action states appear to be consistent across both versions.)

Based on this, we can conditionally check the `action` property. If the last action was a `PUSH`, it’s safe to execute `goBack()`, as the navigation originated from the app itself. Otherwise, we can call `history.push('/')` to perform the default redirect to the correct board.

Let me know what you think or if you have any suggestions!

zkarmi: @RodriSanchez1 Thank you for that insight!

I've tested some cases and these are my findings (recording below):

- Case 1: If a user comes to the site for the first time, and somehow lands initially on `/login-signup`, closing with the X icon will leave them on the WelcomeScreen, which seems confusing. This is because `isFirstVisit && !isLogged` is `true`, so `/` is the WelcomeScreen component:
https://github.com/cboard-org/cboard/blob/1a7e0a97a963ad42636242cb1ced58d2c83a6a5a/src/components/App/App.component.js#L84-L90
- Case 2: Once a user has been on the Board screen, if they go to the AuthScreen they will get sent back to the most recent Board screen.
- Case 3 & 4: If the user navigates away from the app but somehow gets back directly to the AuthScreen, Close will again bring them to the most recent boards screen. This is because `history.action === 'POP'`, which will cause `history.push('/')`, and the Route shown above renders `BoardContainer`.
- Case 4 (not in video): I have tested that reaching AuthScreen from the Settings page sends you back to Settings if you close it.

https://github.com/user-attachments/assets/b8f9e662-62dc-4339-8862-a3711cb9a023

Case 1 seems like it would be uncommon, but not a great UX. Do we want to handle it differently?
Are the rest OK?

RodriSanchez1: Awesome, great job!

Yeah, Step 1 is definitely a bit unusual — I was already aware of that behavior. It's okay for now, since `/login-signup` isn't the most common entry point to Cboard.

I’ve been able to reproduce the issue when clearing the cache and reloading any URL other than `/`, typically something like `/board/root`. If you then go to Settings and and close it, you're redirected to `/`, and the welcome screen is shown — that’s the Step 1 behavior you mentioned, which is slightly different.

Let’s create an issue for this and fix it in a separate PR.

RodriSanchez1: When you have time, please commit the changes

zkarmi: @RodriSanchez1 committed! I also created a new issue here: #1949

RodriSanchez1 (approved):
Awesome! Good work!

PR #1946, opened 2025-06-20
Improve userData handling and clean up Language Redux actions

### LINKED ISSUE #1933: Avoid clean user data if update user is called with an empty payload
Inside the getKeysFromApiUserDataResponse dispatched in this [line](https://github.com/cboard-org/cboard/blob/2824ecf9f7749a5fe55a5af39dce66b2675663ac/src/components/App/App.reducer.js#L66) we are setting the user in the store as a empty object {} if updateUser is called with an undefined payload.
This cause the lost of the user data in the Redux Store when updateUser is called without params.
We can set the already user from the store instead of an empty object to avoid removing the user

### LINKED ISSUE #1934: Avoid call updateUser on changeLang with `undefined` param
Inside our changeLang action, we are calling updateUser without params. This make no sense because it is setting the redux store userData to `undefined` 
https://github.com/cboard-org/cboard/blob/2824ecf9f7749a5fe55a5af39dce66b2675663ac/src/providers/LanguageProvider/LanguageProvider.actions.js#L9 
This make issues on our app in places that the `userData` is used to know if the user is logged in or no.

tomivm:
- Retain existing userData state when no new data is provided for updateUser reducer. Close #1933 
- Removes a line where a Redux action creator was called without being dispatched. Calling an action creator without dispatch does not affect the Redux store or trigger any side effects, as action creators are pure functions that only return action objects or thunks. Removing this unused call improves code clarity and eliminates unnecessary function executions. Close #1934

### DIFF
--- a/src/components/App/App.reducer.js
+++ b/src/components/App/App.reducer.js
@@ -167,7 +167,7 @@ function appReducer(state = initialState, action) {
     case UPDATE_USER_DATA:
       return {
         ...state,
-        userData: getKeysFromApiUserDataResponse(action.userData)
+        userData: action.userData ? getKeysFromApiUserDataResponse(action.userData) : state.userData
       };
     case SET_UNLOGGED_USER_LOCATION:
       return {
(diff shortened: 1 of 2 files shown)


RodriSanchez1 (approved):

RodriSanchez1:
awesome!

PR #1948, opened 2025-06-20
Refactor login to object component

### LINKED ISSUE #1847: "Wrong email or password" error.
# How To Reproduce

1. Navigate to the Login Screen
2. Tap the Login Button
3. In the Login Dialogue:
- Enter a valid email address in the "Email" field. 
- Enter an incorrect password in the "Password" field.
4. Tap the Login button.
5. Observe the system displaying the error message: "Wrong email or password".
6. Tap the cancel button, and the dialogue dismisses.
7. Tap the Login Button again.
8. The system shows the dialogue message "Wrong email or password"

**Expected Result:**

The "Wrong email or password" error message should not be displayed again immediately. 

The application should either:
Attempt a new login with the existing (incorrect) credentials.
Clear the error state and allow the user to enter new credentials without the error message reappearing.
Actual Result:

The "Wrong email or password" error message is displayed again immediately, even though the user has not yet entered new credentials.

Image

### LINKED ISSUE #1940: Refactor Login to functional components
1. Convert the class component `Login.component.js` to a functional component
2. Double check the unit test of the `Login.component.test.js`.

gonojuarez:
close #1940 
close #1847

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "react-joyride": "^2.5.4",
     "react-markdown": "^5.0.3",
     "react-media-recorder": "^0.7.1",
-    "react-redux": "^9.2.2",
+    "react-redux": "^5.1.2",
     "react-router-dom": "^5.3.4",
     "react-scannable": "0.0.18",
     "react-share": "^2.3.1",
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "react-joyride": "^2.5.4",
     "react-markdown": "^5.0.3",
     "react-media-recorder": "^0.7.1",
-    "react-redux": "^5.1.2",
+    "react-redux": "^9.2.2",
     "react-router-dom": "^5.3.4",
     "react-scannable": "0.0.18",
     "react-share": "^2.3.1",
--- a/src/components/Account/Login/Login.component.js
+++ b/src/components/Account/Login/Login.component.js
@@ -1,7 +1,6 @@
-import React, { Component } from 'react';
-import PropTypes from 'prop-types';
+import React, { useState } from 'react';
 import { connect } from 'react-redux';
-import { FormattedMessage, injectIntl, intlShape } from 'react-intl';
+import { FormattedMessage, injectIntl } from 'react-intl';
 import { Formik } from 'formik';
 import classNames from 'classnames';
 import Typography from '@material-ui/core/Typography';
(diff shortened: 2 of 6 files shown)


gonojuarez:
@zkarmi can you chek it

src/components/Account/Login/Login.component.js
```diff
@@ -18,127 +17,127 @@ import messages from './Login.messages';
...
+    },
+    [prevIsDialogOpen, isDialogOpen]
+  );
 
-    this.setState({
-      isLogging: true,
-      loginStatus: {}
-    });
+  function usePrevious(value) {
+    const ref = React.useRef();
```
zkarmi: Since you are importing `useState` and `useEffect` directly, Is there a reason not to do the same for `useRef`?

src/components/Account/Login/Login.component.js
```diff
@@ -18,127 +17,127 @@ import messages from './Login.messages';
...
+      }
+    },
+    [prevIsDialogOpen, isDialogOpen]
+  );
 
-    this.setState({
-      isLogging: true,
-      loginStatus: {}
-    });
+  function usePrevious(value) {
```
zkarmi: This custom hook will be needed for all 3 modals (and perhaps elsewhere on the app?). Also, defining it within the component means it will be redefined on every re-render (Edit: probably not an issue?). 
I'd recommend extracting it to a file the other components can reference, or at the least defining it outside the component function.

src/components/Account/Login/Login.component.js
```diff
@@ -18,127 +17,127 @@ import messages from './Login.messages';
...
-  handleSubmit = values => {
-    const { login } = this.props;
+  useEffect(
+    () => {
+      if (isDialogOpen && isDialogOpen !== prevIsDialogOpen) {
+        setLoginStatus({});
+        setIsLogging(false);
+      }
+    },
+    [prevIsDialogOpen, isDialogOpen]
```
zkarmi: `prevIsDialogOpen` would only change after a re-render that follows `isDialogOpen` changing. Including it in the dependency array would cause an unnecessary extra useEffect run that will always result in a no-op (because it'd essentially be "catching up" to `isDialogOpen`, meaning the condition `isDialogOpen !== prevIsDialogOpen` would always be `false`).

src/components/Account/Login/Login.component.js
```diff
@@ -18,127 +17,127 @@ import messages from './Login.messages';
...
+          onSubmit={handleSubmit}
+          validationSchema={validationSchema}
+        >
+          {({ values, errors, handleChange, handleSubmit }) => (
+            <form className="Login__form" onSubmit={handleSubmit}>
+              <TextField
+                error={errors.email}
+                label={intl.formatMessage(messages.email)}
+                name="email"
+                values={values.email}
```
zkarmi: Typo here, should be `value={values.email}`

zkarmi (changes requested):
@gonojuarez My day got away from me, sorry! 😅 I left some comments. I am not familiar with Formik yet, but the changes in that section make sense to me at a high level.

src/components/Account/Login/Login.component.js
gonojuarez: thanks @zkarmi for this comments

src/components/Account/Login/Login.component.js
gonojuarez: thanks @zkarmi

package.json
```diff
@@ -74,7 +74,7 @@
     "react-joyride": "^2.5.4",
     "react-markdown": "^5.0.3",
     "react-media-recorder": "^0.7.1",
-    "react-redux": "^5.1.2",
+    "react-redux": "^9.2.2",
```
RodriSanchez1: Why did you bump the redux library?

zkarmi (approved):
This looks good to me

RodriSanchez1 (approved):
LGTM

PR #1951, opened 2025-06-24
Refactor reset password

### LINKED ISSUE #1847: "Wrong email or password" error.

### LINKED ISSUE #1942: Refactor Resets Password to Functional Components

gonojuarez:
close #1942 
close #1847


src/components/Account/ResetPassword/ResetPassword.component.js
```diff
@@ -18,103 +18,104 @@ import { forgot } from './ResetPassword.actions';
...
+    setIsSending(true);
+    setForgotState({});
 
     forgot(values)
-      .then(res => this.setState({ forgotState: res }))
-      .catch(err => this.setState({ forgotState: err }))
-      .finally(() => this.setState({ isSending: false }));
+      .then(res => setForgotState(res))
+      .catch(err => setForgotState(err))
+      .finally(() => setIsSending(false));
```
RodriSanchez1: use a try catch block here instead. Set the `setIsSending(false)` only in the `catch` block to avoid unexpected updates of an unmonted component

src/components/Account/ResetPassword/ResetPassword.component.js
```diff
@@ -18,103 +18,104 @@ import { forgot } from './ResetPassword.actions';
...
-              </Typography>
-            ) : (
-              <Typography color="inherit">{forgotState.message}</Typography>
+            {({ values, errors, handleChange, handleSubmit }) => (
+              <form className="Forgot__form" onSubmit={handleSubmit}>
+                <TextField
+                  error={errors.email}
+                  label={intl.formatMessage(messages.email)}
+                  name="email"
+                  value={values.email}
```
RodriSanchez1: You don't need to manually set the value property on the input, since <Formik> automatically controls and updates the form values based on its internal state.

src/components/Account/ResetPassword/ResetPassword.component.js
```diff
@@ -18,103 +18,104 @@ import { forgot } from './ResetPassword.actions';
...
+            <Typography color="inherit">
+              {intl.formatMessage(messages.resetPasswordSuccess)}
+            </Typography>
+          ) : (
+            <Typography color="inherit">{forgotState.message}</Typography>
+          )}
+        </div>
+        {forgotState && !forgotState.success && (
+          <Formik
+            initialValues={{ email: '' }}
```
RodriSanchez1: Define the `intialValue` constant outside of the component to avoid the object redefinition on each render

RodriSanchez1 (changes requested):
Good work

src/components/Account/ResetPassword/ResetPassword.component.js
```diff
@@ -18,103 +18,104 @@ import { forgot } from './ResetPassword.actions';
...
+            onSubmit={handleSubmit}
+            validationSchema={validationSchema}
+            enableReinitialize
```
RodriSanchez1: what this change does?

gonojuarez: I will remove it, that reset the formik to initial values

src/components/Account/ResetPassword/ResetPassword.component.js
gonojuarez: In Formik V2 and above, initial values are required because they define the initial state of the form.
If you don't provide them, values will be undefined, and the form won't work well.

src/components/Account/ResetPassword/ResetPassword.component.js
gonojuarez: I Just change it to a try catch block, the setIsSending in finally it is right.
because the app end the try catch  action and change the status to false.

RodriSanchez1 (approved):

PR #1960, opened 2025-06-29
feat(pdf): add optional tile borders to PDF export (#1954)

### LINKED ISSUE #1954: Add optional tiles border on exported PDF
An user report:

If possible, it could be very helpful to have borders for tiles (buttons and folders), especially for tiles with a white background as when Boards are exported as a PDF there are no visible borders compared to what you could see if the tiles had a different background colour. 

For printable Communication Choice Boards, it can be quite common to have borders for different choices (tiles). This can help differentiate the choices and assist the communicator with selecting the choices they would like to.

Image

dhruv-git-sys:
***Add optional tile borders to PDF export (#1954)***

This PR adds an option to show borders around tiles when exporting boards as PDF.

- Adds a checkbox in the Export UI for "Show borders around tiles in PDF".
- Passes the option through the export container and helpers.
- Updates the PDF export code to draw a border around each tile if selected.
- Updates the changelog.

Fixes #1954

**Screenshots (if applicable):**
(Add screenshot of the new checkbox if you have one)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+- [**feature**] Added optional border for tiles when exporting boards as PDF. Users can now toggle "Show borders around tiles in PDF" in the export settings. [#1954](https://github.com/cboard-org/cboard/issues/1954)
+
+
 ## 1.7.1 (23/12/2020)
 
 #### Bug Fixes:
--- a/src/components/Settings/Export/Export.component.js
+++ b/src/components/Settings/Export/Export.component.js
@@ -14,6 +14,9 @@ import InputLabel from '@material-ui/core/InputLabel';
 import FormControl from '@material-ui/core/FormControl';
 import Divider from '@material-ui/core/Divider';
 
+import Checkbox from '@material-ui/core/Checkbox';
+import FormControlLabel from '@material-ui/core/FormControlLabel';
+
 import FullScreenDialog from '../../UI/FullScreenDialog';
 import messages from './Export.messages';
 
@@ -49,7 +52,8 @@ class Export extends React.Component {
       singleBoard: '',
       loadingSingle: false,
       loadingAll: false,
-      boardError: false
+      boardError: false,
+      showTileBorders: false,
     };
   }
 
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+- [**feature**] Added optional border for tiles when exporting boards as PDF. Users can now toggle "Show borders around tiles in PDF" in the export settings. [#1954](https://github.com/cboard-org/cboard/issues/1954)
+
+
```
RodriSanchez1, line 5: Avoid this change please

RodriSanchez1:
This is awesome! I will do the review this week. Thanks a lot

src/components/Settings/Export/Export.component.js
```diff
@@ -357,6 +368,16 @@ class Export extends React.Component {
...
                 </ListItemSecondaryAction>
+                <FormControlLabel
+                  control={
+                      <Checkbox
+                        checked={this.state.showTileBorders}
+                        onChange={this.handleTileBorderToggle}
+                        color="primary"
+                      />
+                    }
+                    label={intl.formatMessage({ defaultMessage: "Show borders around tiles in PDF" })}
```
magush27: Here there is an incorrect i18n message usage. It uses inline message `intl.formatMessage({ defaultMessage: "Show borders around tiles in PDF" })` which causes React Intl error. 
To fix it you should add the message to `Export.messages.js` and reference it like `intl.formatMessage(messages.showTileBorders)` as you can see in this same file other messages use `messages.messageName`

src/components/Settings/Export/Export.helpers.js
```diff
@@ -379,15 +379,25 @@ async function generatePDFBoard(
...
+  showTileBorders = false
 ) {
   const header = {
     absolutePosition: { x: 0, y: 5 },
     text: board.name || '',
     alignment: 'center',
     fontSize: 8
   };
-
+  const tileRect = {
```
magush27: Here there are undefined variables that affect the border logic. The code references tileX, tileY, tileWidth, tileHeight, and canvas variables that don't exist in that scope. This code is placed at the top of `generatePDFBoard()` function, but these variables don't exist there. This will cause a runtime error when trying to export PDFs with borders.

The border logic needs to be implemented differently, likely within the actual tile rendering code in `addTileToGrid()`.

PR #1976, opened 2025-07-30
Fix label position rendering when image is not present

### LINKED ISSUE #1975: Add Support for Text-Only Tiles
**Description:**

I would like to suggest a feature to enhance board customization in Cboard: the ability to create **tiles containing only text**, without any symbol or image.

**Feature Requirements:**

* If a user **does not select a symbol or image** when creating a new tile, the tile should automatically default to a **text-only tile**.
* For **text-only tiles**, the text should be displayed in a **standard font style (not bold)**.

**Why This Matters:**

This feature would enhance accessibility and customization for:

* Users with strong **literacy skills** who prefer reading over symbols.
* Users who rely on **alternative access methods** like eye-gaze or switch scanning, where cleaner layouts and clear text alignment improve usability.

Allowing flexible control over text display makes Cboard more inclusive and adaptable to various communication needs.

Harshit-Mourya:
### Summary
This PR fixes the label alignment issue when the image is missing in the `Symbol` component. Earlier, the label would always appear below regardless of the `labelpos` prop. Now, `Above` and `Below` positions are respected even if no image is provided.

### Fix
Wrapped the image container in a div and conditionally rendered the `` tag. No CSS changes required.

### Linked Issue
Fixes #1975 

### Reviewer
Hi @martinbedouret, this PR addresses the alignment issue for the label when no image is present. Let me know if anything needs to be adjusted.

### DIFF
--- a/src/components/Board/Symbol/Symbol.js
+++ b/src/components/Board/Symbol/Symbol.js
@@ -139,11 +139,11 @@ function Symbol(props) {
         props.labelpos !== 'Hidden' && (
           <Typography className="Symbol__label">{label}</Typography>
         )}
-      {src && (
-        <div className="Symbol__image-container">
-          <img className="Symbol__image" src={src} alt="" />
-        </div>
-      )}
+
+      <div className="Symbol__image-container">
+        {src && <img className="Symbol__image" src={src} alt="" />}
+      </div>
+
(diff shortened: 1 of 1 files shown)


martinbedouret (approved):
good job!

PR #1980, opened 2025-08-07
Fix undesired horizontal scroll caused by icon overflow in the toolbar

### LINKED ISSUE #1925: Undesired horizontal scroll caused by icon overflow in the toolbar
**Description**
The amount of icons in the toolbar is causing layout issues due to insufficient space in small devices, resulting in an undesired horizontal scroll. We should explore introducing a responsive solution such as a hamburger menu to manage the available space more effectively.

**Steps to Reproduce:**
1. On a device with a small screen (e.g., iPhone SE, Android phone), open the AAC board.
2. Tap three times on the lock icon to enter edit mode.
3. Select “Edit ” .
5. Observe that the navigation panel overlaps the grid.

Image

magush27:
Responsive design improvements to the `Navbar` component's CSS, specifically targeting mobile devices. The main change is the addition of a media query to adjust the appearance and alignment of icon buttons and groups when the screen width is 768px or less.

* Added a media query in `Navbar.css` to adjust padding, color, display, and alignment of `.MuiIconButton-root` elements within `.Navbar__group--start` and `.Navbar__group--end` for screens 768px wide or less. Also ensures these groups are centered.

closes #1925

### DIFF
--- a/src/components/Board/Navbar/Navbar.css
+++ b/src/components/Board/Navbar/Navbar.css
@@ -61,18 +61,13 @@
   color: #fff;
 }
 
-@media (max-width: 768px) {
-  .Navbar__group--end .MuiIconButton-root,
-  .Navbar__group--start .MuiIconButton-root {
-    padding: 6px;
-    color: #fff;
-    display: flex;
-    align-items: center;
-    justify-content: center;
-  }
+.Navbar__group {
+  align-items: center;
+}
 
-  .Navbar__group--end,
-  .Navbar__group--start {
-    align-items: center;
+@media (max-width: 380px) {
+  .Navbar__group button,
+  .Navbar__group a {
+    padding-inline: 8px;
   }
 }
--- a/src/components/Board/Navbar/Navbar.css
+++ b/src/components/Board/Navbar/Navbar.css
@@ -60,3 +60,19 @@
 .Navbar .LoginSignUpButton:focus {
   color: #fff;
 }
+
+@media (max-width: 768px) {
+  .Navbar__group--end .MuiIconButton-root,
+  .Navbar__group--start .MuiIconButton-root {
+    padding: 6px;
+    color: #fff;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+  }
+
+  .Navbar__group--end,
+  .Navbar__group--start {
+    align-items: center;
+  }
+}


src/components/Board/Navbar/Navbar.css
```diff
@@ -60,3 +60,19 @@
 .Navbar .LoginSignUpButton:focus {
   color: #fff;
 }
+
+@media (max-width: 768px) {
```
tomivm: Hey Agus! I think we can use a lower max-width because this smaller padding is not needed in mid size devices.

src/components/Board/Navbar/Navbar.css
```diff
@@ -60,3 +60,19 @@
...
+  .Navbar__group--end .MuiIconButton-root,
+  .Navbar__group--start .MuiIconButton-root {
```
tomivm: ☘️  Is generally better to avoid using Material UI classnames to style the components.
Is better to use a custom class name like `Navbar__group--button`.
Also i noticed that using `Navbar__group--end` and `Navbar__group--start` is not necessary.
You can use directly `.Navbar__group .MuiIconButton-root` for your actual code. (or `.Navbar__group .Navbar__group--button` taking this comment)

src/components/Board/Navbar/Navbar.css
```diff
@@ -60,3 +60,19 @@
...
+    padding: 6px;
+    color: #fff;
```
tomivm: I didn't understand why you add a color here? it's necessary for something?

src/components/Board/Navbar/Navbar.css
```diff
@@ -60,3 +60,19 @@
 .Navbar .LoginSignUpButton:focus {
   color: #fff;
 }
+
+@media (max-width: 768px) {
+  .Navbar__group--end .MuiIconButton-root,
```
tomivm: I think you can remove all the css lines under this line if you you use 'padding-inline' instead of 'padding'
see:
```suggestion
  .Navbar__group .MuiIconButton-root {
    padding-inline: 6px;
  }
```
Let me know if i am wrong please :)

magush27:
I took your suggestions and made this changes in the commit:
  * Align items in the navbar for all screen sizes.
  * Adjust padding for buttons and links on screens smaller than 380px to fix the overflow.

tomivm (approved):
Looks good!

PR #1989, opened 2025-08-27
Adds authentication to board report API

martinbedouret:
Ensures only authenticated users can report boards.
This prevents unauthorized access and potential abuse
of the reporting functionality.
Also, disables the report button if the user is not authenticated.


magush27:
I tested it in the frontend without auth and is working as expected, great job!

magush27 (approved):

PR #2013, opened 2025-09-16
chore: refactor OutputContainer to remove extra translation logic 

### LINKED ISSUE #1889: Consider remove the translation logic from the Output.container
We should consider removing the translation logic from the Output.container (Inside getDerivedStateFromProps)  as labels are translated on the output without this.
The only reason to translate the output here is for users that want to have an output and change the language of the whole app to see the translation of the outputBar (Really not nice UX for this haahaha)
Maybe we can add a feature to translate the output with AI for bilingual users.

cmdsreedev:
### Why
 - Removed extra translations performed in `getDerivedStateFromProps` because the output props already have translated strings. 
 -  verified by changing language and adding labels to output

Closes #1889

### DIFF
--- a/src/components/Board/Output/Output.container.js
+++ b/src/components/Board/Output/Output.container.js
@@ -16,18 +16,6 @@ import {
 import { changeOutput, clickOutput, changeLiveMode } from '../Board.actions';
 import SymbolOutput from './SymbolOutput';
 
-function translateOutput(output, intl) {
-  const translatedOutput = output.map(value => {
-    let translatedValue = { ...value };
-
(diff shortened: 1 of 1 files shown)


tomivm:
Hi!! it's looks like it is okey. 
it simplify the extra logic of translations.
I am a bit afraid of lost the the translation when user change the language. I think any user use that to translate output ⚖️ 
I was thinking in add a feature to translate with AI. The most difficult part is users that have a TTS that is not multilingual.
Now @RodriSanchez1 is working hard in improve our TTS features

PR #2014, opened 2025-09-16
Fix #1786: disable scroll for large boards

### LINKED ISSUE #1786: Disable scroll for large boards
Dear CBoard team, I am a speech and langauge therapist in Asia. I support students (e.g. with Cerebral Palsy) who have motor difficulties which leads to them struggling with the scrolling feature on the app vocabulary pages. When I trialled CBoard with these students, the scrolling of the screen made it hard for them to select vocabulary accurately. Is there a way to remove the scrolling feature, and instead access the symbols via pages? (E.g. select a button for 'Next Page' and disabling the scrolling on the screen?) Thank you very much! Your advice will be very appreciated. Best wishes, Dawn : )

mdkhurram0207:
- Disabled scroll buttons for boards taller than viewport.
- Scroll buttons now update dynamically on scroll and window resize.
- Tested locally on large and small boards.

### DIFF
--- a/src/components/ScrollButtons/ScrollButtons.js
+++ b/src/components/ScrollButtons/ScrollButtons.js
@@ -11,38 +11,61 @@ const TOP_OFFSET = 5;
 const ScrollButtons = props => {
   const [isScrollTop, setScrollTop] = useState(false);
   const [isScrollDown, setScrollDown] = useState(false);
+  // State to track if the board is large and scroll should be disabled
+  const [isScrollDisabled, setScrollDisabled] = useState(false);
 
   useEffect(
     () => {
       const bc = props.boardContainer.current;
-      const checkScrollLimits = event => {
+      // Exit early if the container isn't available yet
+      if (!bc) {
+        return;
+      }
+
+      const checkScrollState = () => {
+        // Check if scrolled to the top or bottom
         const scrollTop = bc.scrollTop <= TOP_OFFSET;
         const scrollDown =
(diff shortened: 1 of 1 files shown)


magush27:
Thanks for working on #1786! However, testing revealed a critical logic error that breaks scroll button functionality.

**Issue Found**
The scroll buttons are always disabled when they should be enabled, and vice versa.

**Root Cause**
The logic in `ScrollButtons.js` lines 33-35 is inverted:
```
const isLargeBoard = bc.scrollHeight > window.innerHeight;
setScrollDisabled(isLargeBoard);  // BUG: Disables when board is LARGE!
```

**Current behavior:**
- Large board (needs scrolling) → isScrollDisabled = true → Buttons disabled
- Small board (no scrolling needed) → isScrollDisabled = false → Buttons enabled 

**Expected behavior:**
- Large board (needs scrolling) → Buttons enabled
- Small board (no scrolling needed) → Buttons disabled

**Testing Evidence**
- Master branch: Scroll buttons work correctly
- PR #2014 branch: Scroll buttons always grayed out and non-functional

**Additional Issues**
 1. Wrong comparison: Compares `bc.scrollHeight > window.innerHeight` but should compare against the container's height, not the full window
 2. Inverted logic: The disabled state should be `!isLargeBoard` or the variable should be renamed to isSmallBoard

**Recommendation**
Closing this PR as it introduces a regression. If you'd like to resubmit:
- Fix the inverted logic
- Compare against container height instead of window height
- Add test coverage to prevent this type of regression

Thanks again for the contribution!

PR #2015, opened 2025-09-16
ElevenLabs API client infrastructure

### LINKED ISSUE #1991: Implement ElevenLabs API client infrastructure
Create the foundational API client for ElevenLabs integration, following the existing Azure Speech Services pattern in the codebase.

## Acceptance Criteria

- [ ] Add ElevenLabs API constants to `src/constants.js` (API base URL, timeout settings)
- [ ] Create ElevenLabs API calls in `src/api/api.js` with methods:
  - `getElevenLabsVoices()` - fetch available cloned voices
  - `synthesizeSpeechElevenLabs(text, voiceId, settings)` - TTS synthesis
- [ ] Add API key validation and connection testing
- [ ] Implement error handling for API failures, rate limits, and network issues
- [ ] Add internet connectivity checks before API calls
- [ ] Support for custom API endpoints and regional configurations

## Technical Implementation Notes

```javascript
// Add to src/constants.js
export const ELEVENLABS_API_BASE_URL = 'https://api.elevenlabs.io/v1';
export const ELEVENLABS_DEFAULT_TIMEOUT = 10000;

// Add to src/api/api.js
async getElevenLabsVoices() {
  const apiKey = this.getElevenLabsApiKey();
  if (!apiKey) return [];

  const headers = { 'xi-api-key': apiKey };
  const url = `${ELEVENLABS_API_BASE_URL}/voices`;

  try {
    const { status, data } = await this.axiosInstance.get(url, { headers });
    if (status === 200) return data.voices || [];
    return [];
  } catch (err) {
    console.error('ElevenLabs API error:', err.message);
    return [];
  }
}
```

## Files to Modify

- `src/constants.js`
- `src/api/api.js`
- `src/api/api.test.js` (add tests)

magush27:
Added support for ElevenLabs as an additional text-to-speech provider, allowing users to configure and use their ElevenLabs API key.

**Changes:**
- Added a new `elevenlabs.js` engine class that manages API key initialization, validation, fetching available voices, and synthesizing speech via the ElevenLabs API. Includes error handling for unauthorized keys and rate limits.
- Defined new constants for ElevenLabs API base URL and default timeout in `constants.js`.
- Extended the speech provider's Redux state to store the ElevenLabs API key, with corresponding action (`CHANGE_ELEVENLABS_API_KEY`) and reducer logic to update and persist the key.
- Updated the `Speech.container.js` component to:
  - Initialize the ElevenLabs engine with the stored API key on mount.
  - Provide a handler to update and validate the API key, dispatch changes to Redux, and persist the setting.
  - Include the ElevenLabs API key in the relevant speech settings logic and Redux mappings.
- Added imports and updated references in `Speech.container.js` to use the new ElevenLabs constants and engine. 

Closes #1991


martinbedouret (changes requested):
I disagree that api key is on the local storage as it would suppose you have to configure your key every time you use a new device. cboard is a multi platform application and make not sense you have to configure any device by separate. All settings you configure are shared across all of your devices and this is one more setting.

magush27:
> I disagree that api key is on the local storage as it would suppose you have to configure your key every time you use a new device. cboard is a multi platform application and make not sense you have to configure any device by separate. All settings you configure are shared across all of your devices and this is one more setting.

yes, also the localStorage is very insecure for this purpose.. maybe is better if we use the backend to store the key and get it from there. what do you think?

martinbedouret:
Absolutely Agus, take a look on how we are saving the settings on the back-end and replicate the same mechanism.

src/components/Settings/Speech/Speech.container.js
```diff
@@ -44,7 +46,26 @@ export class SpeechContainer extends Component {
...
-  async componentDidMount() {}
+  async componentDidMount() {
+    const {
+      speech: { elevenLabsApiKey }
+    } = this.props;
+
+    if (elevenLabsApiKey) {
+      elevenLabsEngine.initialize(elevenLabsApiKey);
+    }
+  }
```
RodriSanchez1: Do you really  need to initilize the instance on the did mount?

src/components/Settings/Speech/Speech.container.js
```diff
@@ -115,7 +138,7 @@ export class SpeechContainer extends Component {
       lang,
       speech: {
         voices,
-        options: { voiceURI, pitch, rate }
+        options: { voiceURI, pitch, rate, elevenLabsApiKey }
       }
```
RodriSanchez1, line 135: You can get the `elevenLabsApiKey` directly from the props 

```ts
const  { elevenLabsApiKey } = this.props
```

as you are mapping it in the `mapStateToProps()` at the bottom of the file

RodriSanchez1 (approved):
Good job! @magush27 please update the PR description

src/components/Settings/Speech/Speech.container.js
magush27: I think that without this step, any ElevenLabs TTS features would fail until the engine is properly initialized with the user's API key... maybe I can simplify it like this to avoid redundancy as is called in the `mapStateToProps()`
```
async componentDidMount() {
  const { elevenLabsApiKey } = this.props;
  if (elevenLabsApiKey) {
    elevenLabsEngine.initialize(elevenLabsApiKey);
  }
}
```

RodriSanchez1: You are right! But you should initilize the instance in the `tts.js` file instead.

See this line https://github.com/cboard-org/cboard/blob/8df3cf4077771a5f52dc2d32b605533bf9284675/src/providers/SpeechProvider/tts.js#L73

Thats for the Azure instance, you should do something similar

magush27: oh! right, I'll change the code

magush27:
> Good job! @magush27 please update the PR description
yes! I just updated it

src/api/api.js
```diff
@@ -1,13 +1,14 @@
...
 import { alpha2ToAlpha3T } from '@cospired/i18n-iso-languages';
-
 import {
   API_URL,
   ARASAAC_BASE_PATH_API,
   GLOBALSYMBOLS_BASE_PATH_API,
   AZURE_VOICES_BASE_PATH_API,
-  AZURE_SPEECH_SUBSCR_KEY
+  AZURE_SPEECH_SUBSCR_KEY,
+  ELEVENLABS_API_BASE_URL,
```
martinbedouret: Remove unused constants

src/components/Settings/Speech/Speech.container.js
```diff
@@ -73,14 +87,16 @@ export class SpeechContainer extends Component {
     this.updateSettingsTimeout = setTimeout(async () => {
       const {
         speech: {
-          options: { voiceURI, pitch, rate }
+          options: { voiceURI, pitch, rate },
+          elevenLabsApiKey
```
martinbedouret: I don't think you have to get the `elevenLabsApiKey` from props here as you are passing the values in the function arguments

src/components/Settings/Speech/Speech.container.js
```diff
@@ -73,14 +87,16 @@ export class SpeechContainer extends Component {
...
         }
       } = this.props;
 
       const speech = {
         voiceURI,
         pitch,
         rate,
+        elevenLabsApiKey,
```
martinbedouret, line 99: same comment than above

src/components/Settings/Speech/Speech.container.js
```diff
@@ -115,7 +131,7 @@ export class SpeechContainer extends Component {
       lang,
       speech: {
         voices,
-        options: { voiceURI, pitch, rate }
+        options: { voiceURI, pitch, rate, elevenLabsApiKey }
```
martinbedouret: no no, the `elevenLabsApiKey` is not an element in the options array!

src/providers/SpeechProvider/engine/elevenlabs.js
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    try {
+      const response = await fetch(`${ELEVENLABS_API_BASE_URL}/v2/voices`, {
+        method: 'GET',
+        headers: {
+          'xi-api-key': this.apiKey,
+          'Content-Type': 'application/json'
+        }
+      });
```
martinbedouret: How do you filter just cloned voices?

martinbedouret (changes requested):

src/api/api.js
magush27: oh, yes! I just deleted them

src/components/Settings/Speech/Speech.container.js
magush27: you are right, I need to pass it outside the speech props in this case to use it from `this.props.elevenLabsApiKey` form the `mapStateToProps`

src/components/Settings/Speech/Speech.container.js
magush27: In this case I think is right to pass `elevenLabsApiKey` as is the new settings object that updates the user speech preferences in the api trought updateSettings. The `elevenLabsApiKey` here is being taken from `mapStateProps` as I changed it in the code that is before

src/components/Settings/Speech/Speech.container.js
magush27: yes, this is wrong.. I just changed it

src/providers/SpeechProvider/tts.js
```diff
@@ -50,6 +51,17 @@ const initAzureSynthesizer = () => {
   );
 };
 
+const initElevenLabsEngine = () => {
```
martinbedouret: The `tts.js` is a module. When this module is imported during app startup, the Redux store hasn't been created yet. The `getStore()` function returns undefined because the store initialization happens later in the React app lifecycle. 
You have to Initialize ElevenLabs engine lazily when it's first needed, rather than during module load.
```
// ...existing code...

let elevenLabsInitialized = false;

const initElevenLabsEngine = () => {
  const store = getStore();
  if (!store) {
    // Store not available yet, skip initialization
    return;
  }
  
  const {
    speech: { elevenLabsApiKey }
  } = store.getState();

  if (elevenLabsApiKey) {
    elevenLabsEngine.initialize(elevenLabsApiKey);
    elevenLabsInitialized = true;
  }
};

const ensureElevenLabsInitialized = () => {
  if (!elevenLabsInitialized) {
    initElevenLabsEngine();
  }
};

// ...existing code...

initAzureSynthesizer();
// Remove this line: initElevenLabsEngine();

const tts = {
  // ...existing code...
  
  reinitializeElevenLabs() {
    elevenLabsInitialized = false;
    initElevenLabsEngine();
  },

  async speak(text, options, setCloudSpeakAlertTimeout) {
    // Ensure ElevenLabs is initialized when speak is called
    ensureElevenLabsInitialized();
    
    // ...existing code...
  }

  // ...existing code...
};
```

martinbedouret (changes requested):

src/providers/SpeechProvider/tts.js
magush27: awesome! great observation, I already implemented this in the things i'm going to push

src/providers/SpeechProvider/engine/elevenlabs.js
magush27: I didn't put it yet, to filter just cloned and/or generated voices we need to make a call to `https://api.elevenlabs.io/v2/voices?voice_type=personal` to get just the personal voices, this includes both. Now I'll change it in the code

src/components/Settings/Speech/Speech.container.js
RodriSanchez1: Martin, you are correct that the prop comes in through the params. However, it’s important to note that **all speech-related fields must be taken from the props and passed into the `updateSettings` function**.  

The reason is that the server expects the full set of speech settings. If any field is missing, the server will erase it from the stored configuration. This behavior comes directly from how the function is implemented on the server side:  

https://github.com/cboard-org/cboard-api/blob/21d4f7e26a284943370cbdb1ff35d0785c9ebe79/api/controllers/settings.js#L8-L33

magush27: > Agus you can keep accessing the prop as you are doing now, or use `this.props.elevenLabsApiKey` as you suggested.
> 
> Martin, you are correct that the prop comes in through the params. However, it’s important to note that **all speech-related fields must be taken from the props and passed into the `updateSettings` function**.
> 
> The reason is that the server expects the full set of speech settings. If any field is missing, the server will erase it from the stored configuration. This behavior comes directly from how the function is implemented on the server side:
> 
> https://github.com/cboard-org/cboard-api/blob/21d4f7e26a284943370cbdb1ff35d0785c9ebe79/api/controllers/settings.js#L8-L33

Yes, I changed it to access it from the mapStateProps in the new commits and pass it to the object that updates the backend

src/providers/SpeechProvider/engine/elevenlabs.js
RodriSanchez1: Rename this function to `getElevenLabsPersonalVoices` :)

magush27: great, already changed it in the new commit

magush27:
I made some changes including the suggestions from above

 ### Changes made:

  **`engine/elevenlabs.js`:**
  - Exported `ElevenLabsEngine` class
  - Removed `initialize()` method, initialization now happens through constructor only
  - Made API key required in constructor with explicit validation:
    - Throws error if no API key provided
    - Throws error if API key format is invalid
  - Exported `validateApiKeyFormat()` helper function for use in other modules

  **`Speech.container.js`:**
  - Updated import to use `validateApiKeyFormat` function instead of instance method
  - Validation now happens before Redux state update and engine initialization

  **`tts.js`:**
  - Added `elevenLabsSynthesizer` variable to hold engine instance (similar to `azureSynthesizer`)
  - Updated `initElevenLabsEngine()` to `initElevenLabsSynthetizer()` to instantiate class: `new ElevenLabsEngine(apiKey)`
  - Updated `ensureElevenLabsInitialized()` to check instance existence and initialization state
  - Updated `reinitializeElevenLabs()` to reset instance properly

I'm still not doing further `tts.js` configurations regarding `speak()` as this is stuff from another issue

martinbedouret (approved):

src/components/Settings/Speech/Speech.container.js
```diff
@@ -46,6 +49,17 @@ export class SpeechContainer extends Component {
 
   async componentDidMount() {}
 
+  handleUpdateElevenLabsApiKey = async apiKey => {
+    const { changeElevenLabsApiKey } = this.props;
+
+    if (apiKey && !validateApiKeyFormat(apiKey)) {
+      throw new Error('Invalid API key format');
+    }
```
RodriSanchez1, line 57: Need to be sure that the function that calls the `handleUpdateElevenLabsApiKey` is wrapped within a try catch. If not could break the app

src/providers/SpeechProvider/engine/elevenlabs.js
```diff
@@ -0,0 +1,106 @@
...
+      throw new Error('Invalid ElevenLabs API key format');
+    }
+    this.apiKey = apiKey;
+  }
+
+  isInitialized() {
+    return this.apiKey !== null;
+  }
+
+  async getElevenLabsPersonalVoices() {
```
RodriSanchez1, line 23: It can be renamed to `getPersonalVoices` to simplify

src/providers/SpeechProvider/engine/elevenlabs.js
```diff
@@ -0,0 +1,106 @@
...
+
+      const data = await response.json();
+      return data.voices || [];
+    } catch (error) {
+      console.error('ElevenLabs API error:', error.message);
+      return [];
+    }
+  }
+
+  async synthesizeSpeechElevenLabs(text, voiceId, settings = {}) {
```
RodriSanchez1, line 52: It can be renamed to `synthesizeSpeech`

RodriSanchez1:
Good work @magush27  ! congrats

Issue #1979, opened 2025-08-05
Relocate Rows and Columns Selector Controls from Grid Overlay


RodriSanchez1:
Currently, the row and column selector controls (arrows) are placed directly over the grid, which can interfere with user interaction and visibility.

### Proposal
Based on a recent suggestion, here are three proposed options for relocating or redesigning these controls:

1. **Option 1**: Place the control arrows as a new row and column, visually similar to the existing rows/columns (see the image for an example with row controls; column controls would be similar but vertical).

<img width="320" height="568" alt="Image" src="https://github.com/user-attachments/assets/1a9cefbd-24d4-481f-adef-b5b480324fe6" />

2. **Option 2**: Integrate the row and column controls into the existing edit options for the grid.

 <img width="320" height="568" alt="Image" src="https://github.com/user-attachments/assets/746736da-c969-452a-aa52-98d3087a33fa" />

3. **Option 3**: Similar to option 2, but place the controls at the bottom of the screen.

 <img width="320" height="568" alt="Image" src="https://github.com/user-attachments/assets/8156d457-f2d3-46b2-a203-38b47770e43d" />

Material UI 4 components (buttons and text) are available for these UI elements.

---
_Reference: Originally posted by @magush27 in [#1925](https://github.com/cboard-org/cboard/issues/1925#issuecomment-3151231556)_

Priyanshu1-62:
We need to run both cboard and cboard-api for local development. The backend requires a few necessary API keys and secrets to function, especially SendGrid API key. I can short-circuit authentication functions to bypass email verification checks, but will require me to understand the authentication flow and file structures !!. Is this the expected way do locally run the app?

RodriSanchez1:
Hi! You dont need to run the `cboard-api` to work on this issue. Just use the app without log in. 
Steps to reproduce open the image view:
- 4 clicks on the clock
- Click Edit
- Select the Fixed switch to convert the board to fixed
- The Rows and Collumns selector will be showed

Let me know if you need further assist or if you want to be assigned on this issue

Priyanshu1-62:
I would like to be assigned to this issue. Thanks for the opportunity! I plan to implement Option 2 for relocating & redesigning the controls.

RodriSanchez1:
Okay! Nice! Yes, I also like the option 2

Priyanshu1-62:
Hii @RodriSanchez1 , I have completed working on the issue and it works totally fine in local environment. 
[...]

PR #2027, opened 2025-10-03
Relocated row and column selectors

realasuarez:


magush27:
@realasuarez can you make a more complete description on what you are doing in this PR so I can test it accordingly? Thanks for your contribution!

magush27:
This is a proposal to solve the issue #1979 . 
It has a problem that is it has hardcoded positioning in mobile and desktop and the fixed pixel values don't scale across different screen sizes and resolutions. Also, it doesn't account for toolbar variations or other UI configs. It may cause overlapping with toolbar elements, smaller screens or different layouts.
It also has a counter-intuitive button logic the down arrow removes a row and the up arrow adds a row but in typical user experiences the down arrow suggests adding to the bottom and the up arrow suggests removing from top..
Other problem I found is that it removes the dynamic position feature, the original code dynamically positioned column buttons based on navigation settings. This PR removes that adaptability, when big scroll buttons are active it may cause overlap because of the fixed positioning. 
This is the way the implementation looks like:
imagen

RodriSanchez1:
As the PR does not resolve the inteded issue I will be closing it .  Thanks @realasuarez for your contribution

PR #2030, opened 2025-10-12
Fix: Prevent Language Settings freeze when no TTS voices are available (#1923)

### LINKED ISSUE #2026: Language settings page freezes on Linux desktop when no TTS voices are available
**Description:**
A user reported that on Linux desktop machines, the language settings page in CBoard freezes if there are no Text-to-Speech (TTS) voices available.

Even if the browser’s Text-to-Speech relies on an OS-level TTS service that might not be present, CBoard should not hang. Instead, it should show an empty list of voices or provide a fallback behavior.

**Steps to Reproduce:**
1. Run CBoard on a Linux desktop system without any OS-level TTS voices installed.
2.  Navigate to the Language Settings page.
3. Observe that the page freezes instead of showing an empty list.

**Expected Behavior:**
- The page should not freeze.
- If no TTS voices are available, the language list should be empty or fallback behavior should be triggered gracefully.

**Actual Behavior:**
- The Language Settings page freezes, blocking further interaction.

**Environment:**
- Platform: Linux desktop (version unknown)
- Browser: Firefox

**Additional Notes:**

The user asked if CBoard already has test coverage for Linux desktops. They expressed willingness to add or extend existing tests if none currently exist.

sharan-deep21:
Description:
This PR fixes an issue where the Language Settings page on Linux desktop would freeze if no Text-to-Speech (TTS) voices were available from the OS or browser.

 Root Cause:
The app’s `getVoices()` function assumed that `speechSynthesis.getVoices()` would always return a list.  
On some Linux systems, it returns an empty array or throws an error, causing the page to hang while awaiting results.

 Fix Implemented:
- Added safety handling in `src/providers/SpeechProvider/tts.js`:
  - If `speechSynthesis.getVoices()` returns an empty array or throws, the function now gracefully resolves with an empty list instead of freezing.
- Added console log to simulate voice unavailability for testing.

 Testing Done:
✅ Verified on Linux:
- No freeze when `getVoices()` returns empty.
- Language Settings page loads normally with an empty list.
- Console shows expected log: “✅ Simulated no voices available”.

### Issue Reference
Fixes #2026

### DIFF
--- a/src/components/Communicator/CommunicatorDialog/CommunicatorDialogBoardItem.component.js
+++ b/src/components/Communicator/CommunicatorDialog/CommunicatorDialogBoardItem.component.js
@@ -772,46 +772,23 @@ class CommunicatorDialogBoardItem extends React.Component {
             {moment(board.lastEdited).format('DD/MM/YYYY')}
           </div>
 
-          <div
-            className="CommunicatorDialog__boards__item__data__extra"
-            style={{ display: 'flex', alignItems: 'center', gap: '6px' }}
-          >
+          <div className="CommunicatorDialog__boards__item__data__extra">
             {board.isPublic && (
               <Tooltip
                 title={intl.formatMessage(messages.publicBoard)}
                 name="CommunicatorDialog__PropertyOption"
               >
-                <div
-                  style={{ display: 'flex', alignItems: 'center', gap: '4px' }}
-                >
-                  <PublicIcon />
-                  {board.language && (
-                    <span style={{ fontSize: '0.85rem', opacity: 0.8 }}>
-                      {board.language}
-                    </span>
-                  )}
-                </div>
+                <PublicIcon />
               </Tooltip>
             )}
-
             {!board.isPublic && (
               <Tooltip
                 title={intl.formatMessage(messages.privateBoard)}
                 name="CommunicatorDialog__PropertyOption"
               >
-                <div
-                  style={{ display: 'flex', alignItems: 'center', gap: '4px' }}
-                >
-                  <KeyIcon />
-                  {board.language && (
-                    <span style={{ fontSize: '0.85rem', opacity: 0.8 }}>
-                      {board.language}
-                    </span>
-                  )}
-                </div>
+                <KeyIcon />
               </Tooltip>
             )}
-
             {communicator.rootBoard === board.id && (
               <Tooltip
                 title={intl.formatMessage(messages.rootBoard)}
@@ -820,13 +797,12 @@ class CommunicatorDialogBoardItem extends React.Component {
                 <HomeIcon />
               </Tooltip>
             )}
-
             {activeBoardId === board.id && (
               <Tooltip
                 title={intl.formatMessage(messages.activeBoard)}
                 name="CommunicatorDialog__PropertyOption"
               >
-                <EditIcon />
+                <RemoveRedEyeIcon />
               </Tooltip>
             )}
           </div>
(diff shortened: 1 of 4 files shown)


src/providers/SpeechProvider/tts.js
```diff
@@ -197,19 +197,23 @@ const tts = {
       }
 
       // Android
+      // ✅ Fix for Linux freeze when no TTS voices available
```
tomivm: Please avoid leave comments in the code

src/providers/SpeechProvider/tts.js
```diff
@@ -197,19 +197,23 @@ const tts = {
...
+          // 🛠️ Handle case: no voices found
+          if (!voices || voices.length === 0) {
+            console.warn('No TTS voices available. Resolving with empty list.');
             synth.removeEventListener('voiceschanged', voiceslst);
-            // On Cordova, voice results are under `._list`
-            platformVoices = voices._list || voices;
-            resolve(
-              platformVoices.concat(cloudVoices).concat(elevenLabsVoices)
-            );
+            resolve([]); // resolve with empty array instead of freezing
```
tomivm: Instead of resolve with an empty array, we can return the concatenation of the online voices `cloudVoices` and `elevenLabsVoices`

src/providers/SpeechProvider/tts.js
```diff
@@ -197,19 +197,23 @@ const tts = {
...
+            return;
           }
+
+          // 🧠 Normal case: voices found
+          synth.removeEventListener('voiceschanged', voiceslst);
```
tomivm: We can avoid repeat this line in :208 and :214 by moving it to line 204 before the `if (!voices || voices.length === 0)` statement

src/components/Communicator/CommunicatorDialog/CommunicatorDialogBoardItem.component.js
```diff
@@ -772,23 +772,46 @@ class CommunicatorDialogBoardItem extends React.Component {
...
             {board.isPublic && (
               <Tooltip
                 title={intl.formatMessage(messages.publicBoard)}
                 name="CommunicatorDialog__PropertyOption"
               >
-                <PublicIcon />
+                <div
+                  style={{ display: 'flex', alignItems: 'center', gap: '4px' }}
+                >
+                  <PublicIcon />
```
tomivm: The changes in this file are for this PR?

tomivm (changes requested):
Looks Good 💯 . Thanks for collaborate! 
I leave some comments with details that we can improve. 🥇

src/providers/SpeechProvider/tts.js
sharan-deep21: Yeah updated to return concatenated cloud and elevenLabs voices.

src/providers/SpeechProvider/tts.js
sharan-deep21: Thanks for the suggestion! I moved synth.removeEventListener('voiceschanged', voiceslst); above the voice length check to eliminate duplication as recommended.

src/providers/SpeechProvider/tts.js
sharan-deep21: yeah done.

src/components/Communicator/CommunicatorDialog/CommunicatorDialogBoardItem.component.js
sharan-deep21: Fixed it✔, once check it and inform me for any further changes 😊.

tomivm:
@sharan-deep21 Hey :) is ready for review?

PR #2046, opened 2025-11-05
Logout functionality to clear ElevenLabs settings and voices

### LINKED ISSUE #2041: ElevenLabs API key persists in Speech settings after user logout
## Description

When a registered user logs out of **Cboard**, the **ElevenLabs API key** remains visible in `Settings → Speech`.
This key should be cleared automatically upon logout to prevent potential security or privacy issues.

---

## 🧩 Steps to Reproduce

1. Log in to **Cboard** with an account that has a valid ElevenLabs API key configured.
2. Go to `Settings → Speech` and confirm the key is present.
3. Log out of the account.
4. Go back to `Settings → Speech`.

---

## ✅ Expected Behavior

After logout, the **ElevenLabs API key** should be **removed** from local storage and no longer displayed in the Speech settings.

---

## ❌ Actual Behavior

The **API key** persists in the settings even after the user logs out.

https://github.com/user-attachments/assets/fa772cef-4276-4472-a643-959b1142f3a6

---

## 🧪 Environment

* **App version:** `v1.38.0` (current)
* **Platform:** Web / Android / iOS / Windows
* **Browser:** Chrome / Edge / Safari / Firefox

---

## 📝 Additional Notes

* The key persists until manually removed or overwritten by another login.
* Potential **privacy/security risk** if multiple users share the same device.

---

magush27:
Added logic to the `speechProviderReducer` to handle user logout events by cleaning up state related to ElevenLabs voices and settings. This ensures that sensitive information and cached data are cleared when a user logs out.

Changes:

* Added a new case for the `LOGOUT` action in `speechProviderReducer` that:
  * Clears the `elevenLabsApiKey` and resets the `elevenLabsCache` and `elevenLabsVoiceSettings`.
  * Removes all ElevenLabs voices from the `voices` array.
  * Resets ElevenLabs-specific options and clears the selected voice if it was an ElevenLabs voice.
* Updated imports in `SpeechProvider.reducer.js` to include the `LOGOUT` constant from `Login.constants`.

Closes #2041

### DIFF
--- a/src/providers/SpeechProvider/SpeechProvider.reducer.js
+++ b/src/providers/SpeechProvider/SpeechProvider.reducer.js
@@ -26,7 +26,10 @@ import {
   standardizeLanguageCode
 } from '../../i18n';
 import { CHANGE_LANG } from '../LanguageProvider/LanguageProvider.constants';
-import { LOGIN_SUCCESS } from '../../components/Account/Login/Login.constants';
+import {
+  LOGIN_SUCCESS,
+  LOGOUT
+} from '../../components/Account/Login/Login.constants';
 import { DEFAULT_LANG } from '../../components/App/App.constants';
 
 const initialState = {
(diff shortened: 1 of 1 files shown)


src/providers/SpeechProvider/SpeechProvider.reducer.js
```diff
@@ -335,6 +338,37 @@ function speechProviderReducer(state = initialState, action) {
...
+        elevenLabsCache: {
+          voices: [],
+          timestamp: null,
+          ttl: 24 * 60 * 60 * 1000
+        },
+        elevenLabsVoiceSettings: {},
+        voices: nonElevenLabsVoices,
+        options: {
+          ...state.options,
+          voiceURI: isCurrentVoiceElevenLabs ? null : state.options.voiceURI,
```
RodriSanchez1: We should set a voice as default, it cannot be `null`

magush27: right!, in the next commit I fix it to have a fallback to a voice similar to `CHANGE_LANG` one

RodriSanchez1 (approved):

PR #2048, opened 2025-11-10
fix: prevent crash when tile is missing by adding null safety to resolveTileLabel

sharan-deep21:
Fixes
Prevents app crash when a board contains a tile with missing data by adding a null safety check in `resolveTileLabel`.

Details
- Added a check to ensure tile is not null before destructuring.
- Cleans up previous comment.

Tested
- Ran locally, no crash opening boards with missing tile IDs.

Ready for review! @tomivm 
For any further changes required please mention them 😊

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -57,7 +57,13 @@ export const resolveBoardName = ({ name, nameKey }, intl) => {
   return '';
 };
 
-export const resolveTileLabel = ({ label, labelKey }, intl) => {
+export const resolveTileLabel = (tile, intl) => {
+  if (!tile) {
+    return '';
+  }
+
+  const { label, labelKey } = tile;
+
   if (label) return label;
   if (labelKey && intl?.messages?.[labelKey]) {
     return intl.formatMessage({ id: labelKey });


magush27:
I just tested this defensive solution though the navigator console and it's working as expected, it returns an empty string without crashing.

This is the test I've done in the console:
```
 const oldResolveTileLabel = (tile, intl) => {
    const { label, labelKey } = tile;  // Crashes if tile is null
    if (label) return label;
    if (labelKey && intl?.messages?.[labelKey]) {
      return intl.formatMessage({ id: labelKey });
    }
    return '';
  };

  // Test it with null (should crash)
  try {
    oldResolveTileLabel(null, {});
    console.log('OLD: No crash (unexpected)');
  } catch (e) {
    console.log('OLD: Crashed as expected:', e.message);
  }

  // Simulate the new code (with null check) that should not crash
  const newResolveTileLabel = (tile, intl) => {
    if (!tile) {                        // New null check
      return '';
    }
    const { label, labelKey } = tile;
    if (label) return label;
    if (labelKey && intl?.messages?.[labelKey]) {
      return intl.formatMessage({ id: labelKey });
    }
    return '';
  };

  // Test it with null - should NOT crash
  try {
    const result = newResolveTileLabel(null, {});
    console.log('NEW: Returned empty string without crash:', result === '');
  } catch (e) {
    console.log('NEW: Crashed (unexpected):', e.message);
  }
```

It returned:
```
OLD: Crashed as expected: tile is null
NEW: Returned empty string without crash: true
````

So it's working as expected @tomivm

tomivm (approved):

PR #2097, opened 2026-01-22
Implement board synchronization functionality

### LINKED ISSUE #2096: Implement syncBoards function for board synchronization
## Problem
Currently, boards don't have a proper sync mechanism like communicators do:

1. **`syncCommunicators`** (existing): Compares local vs remote communicators by `lastEdited`, pushes local changes if newer, or updates local state if remote is newer
2. **Boards**: Only pull from server via `getApiMyBoards()`, but local changes are not automatically pushed unless explicitly triggered

The `reconcileBoards` function already exists in `Board.reducer.js` but is only used passively in the `GET_API_MY_BOARDS_SUCCESS` reducer case.

## Proposed Solution
Create a `syncBoards` async function that:

1. **Takes remote boards** as parameter (from `getApiMyBoards()` response)
2. **Iterates through remote boards** and compares with local boards
3. **Uses `reconcileBoards` logic** to determine which version is newer based on `lastEdited` timestamp
4. **If local is newer**: Push to server using `updateApiBoard()` or `createApiBoard()`
5. **If remote is newer**: Update local state
6. **For new remote boards**: Add to local state
7. **Dispatch a `SYNC_BOARDS` action** with the reconciled boards array

## Reference Implementation

```javascript
// Communicator.actions.js:309-394 - syncCommunicators pattern to follow
export function syncCommunicators(remoteCommunicators) {
  const reconcileCommunicators = (local, remote) => {
    if (local.lastEdited && remote.lastEdited) {
      if (moment(local.lastEdited).isAfter(remote.lastEdited)) {
        return local;
      }
      // ... reconciliation logic
    }
    return local;
  };

  return async (dispatch, getState) => {
    // ... sync logic that pushes local if newer
  };
}
```

## Files to Modify

- **`src/components/Board/Board.actions.js`**: Add `syncBoards()` function
- **`src/components/Board/Board.constants.js`**: Add `SYNC_BOARDS` constant
- **`src/components/Board/Board.reducer.js`**: Add `SYNC_BOARDS` case (can reuse existing `reconcileBoards` function)

## Acceptance Criteria

- [ ] `syncBoards(remoteBoards)` function created following `syncCommunicators` pattern
- [ ] Local boards with newer `lastEdited` are pushed to server
- [ ] Remote boards with newer `lastEdited` update local state
- [ ] New remote boards are added to local state
- [ ] `SYNC_BOARDS` action dispatched with reconciled boards
- [ ] Unit tests added for `syncBoards` function
- [ ] Function is called from appropriate location (e.g., after `getApiMyBoards()`)

## Related Code

| Function | Location |
|----------|----------|
| `syncCommunicators` | `src/components/Communicator/Communicator.actions.js:309-394` |
| `reconcileBoards` | `src/components/Board/Board.reducer.js:62-72` |
| `updateApiMarkedBoards` | `src/components/Board/Board.actions.js:760-817` |

tomivm:
Enhance board synchronization by implementing the `syncBoards` action, updating the reducer to manage sync status, and refining board identification logic. This update improves the handling of locally created boards and integrates detailed synchronization states. Tests have been adjusted to ensure accuracy in board classification and synchronization processes.

NOTE: The original descriptionwas EDITED

Close #2096


tomivm:
I use Claude Code to review the changes and I get: 
  1. ADD_BOARDS missing syncStatus (Board.reducer.js:134-138)
    - Boards from server are added without syncStatus, causing inconsistent behavior
    - Should default to SYNC_STATUS.SYNCED for remote boards
  2. Incorrect axios error property (Board.actions.js:638, 685)
    - Uses e.status instead of e.response?.status
    - 404 errors won't be properly detected
  3. Dead/incomplete code in applyRemoteChangesToState (Board.actions.js:629-643)
    - Commented-out dispatch logic
    - No handling when API.getBoard() succeeds

  Important Issues (Should Fix)

  1. Missing tests for syncBoards and applyRemoteChangesToState - the core sync logic
  2. Sequential API calls could be parallelized with Promise.allSettled (not sure maybe intentional)

  What's Done Well

  - Clear PULL-then-PUSH architecture
  - Good test coverage (111 tests pass)
  - Proper syncStatus and isSyncing state tracking
  - Well-documented functions with JSDoc
  - Clean selectors with tests

  Recommendations

  1. Fix axios error property access
  2. Handle syncStatus in ADD_BOARDS
  3. Clean up the commented code block
  4. Add the missing orchestration tests

src/components/Board/Board.actions.js
```diff
@@ -524,19 +531,209 @@ export function downloadImageFailure(message) {
...
+  return { type: SYNC_BOARDS_FAILURE, error: error.message || error };
+}
+
+/**
+ * PULL: Apply remote changes to local Redux state.
+ * Adds new remote boards and updates boards where remote is newer.
+ */
+export function applyRemoteChangesToState(
+  newRemoteBoards,
+  remoteNewerBoards,
```
tomivm: this names are confusing me

src/components/Board/Board.actions.js
```diff
@@ -524,19 +531,209 @@ export function downloadImageFailure(message) {
...
+
+  for (const remote of remoteBoards) {
+    const local = localBoards.find(b => b.id === remote.id);
+
+    if (!local) {
+      newRemote.push(remote);
+      continue;
+    }
+
+    if (moment(remote.lastEdited).isAfter(local.lastEdited)) {
```
tomivm, line 576: what happen if local is newer ?

src/components/Board/Board.actions.js
```diff
@@ -524,19 +531,209 @@ export function downloadImageFailure(message) {
...
+        remoteNewerBoards,
+        remoteDeletedBoardIds
+      } = classifyRemoteBoards(localBoards, remoteBoards);
+
+      // 2. PULL: Apply remote changes to local state (includes remote deletions)
+      await dispatch(
+        applyRemoteChangesToState(
+          newRemoteBoards,
+          remoteNewerBoards,
+          remoteDeletedBoardIds
```
tomivm: i think would be great to use a object for params of this function. same for `classifyRemoteBoards`.

src/components/Board/Board.actions.js
RodriSanchez1: Do you have any suggestion?

src/components/Board/Board.actions.js
RodriSanchez1: If the local board is newer will be pushed to the api as it gonna have the `syncStatus: pending` state

src/components/Board/Board.actions.js
RodriSanchez1: Thats a good idea

RodriSanchez1:
# Sync Follow-Up Issues

Issues to address in future PRs after the initial `syncStatus` implementation.

---

## 1. Handle 409 errors to prevent infinite retry loop

**Problem:** When a board push fails with 409 (server validation error), the board stays `PENDING` forever. The user permanently sees "unsynced changes" with no way to resolve it.

**Affected code:**
- `pushLocalChangesToApi` catch block — [Board.actions.js:717](src/components/Board/Board.actions.js#L717)

**Expected behavior:**
- 409 (validation rejection): Stop retrying. Pull server version if available, mark `SYNCED`
- Transient errors (500, network): Keep `PENDING` for natural retry on next sync

---

## 2. Boards created while logged out get syncStatus: PENDING

**Problem:** `CREATE_BOARD` reducer unconditionally sets `syncStatus: PENDING`. These boards have `email: ''` which is now correctly blocked by the push filter. However, they remain stuck as `PENDING` locally with no way to resolve — the user sees "unsynced changes" even though the boards will never be pushed.

**Affected code:**
- `CREATE_BOARD` reducer — [Board.reducer.js:262](src/components/Board/Board.reducer.js#L262)

**Possible solutions:**
- Don't set syncStatus in `CREATE_BOARD` when not logged in (pass auth context to action)
- On login, adopt orphaned PENDING boards (set their email to the logged-in user)
- On login, clear syncStatus from boards with empty email

---

## 3. Tile operations don't mark the parent board as PENDING

**Problem:** `CREATE_TILE`, `DELETE_TILES`, and `EDIT_TILES` modify a board's tiles but do NOT set `syncStatus: PENDING` on the parent board. For logged-in users this is handled by `updateIfFeaturedBoard()` in the container which calls `updateBoard()`. However, for logged-out users, tile changes modify the board content without entering the sync system — so if the user later logs in, those tile changes are invisible to sync and could be overwritten by the server version on the next pull.

**Affected code:**
- `tileReducer` — [Board.reducer.js:78-101](src/components/Board/Board.reducer.js#L78-L101)
- `CREATE_TILE` / `DELETE_TILES` / `EDIT_TILES` cases — [Board.reducer.js:284-304](src/components/Board/Board.reducer.js#L284-L304)

**Related to:** Issue #2 and #4 — part of the broader unlogged user modifications problem.

---

## 4. Identity boundary on login

**Problem:** When a user with an existing account logs in on a device that has local modifications, those changes could conflict with their server state.

**Expected behavior:**
- First-time signup: local changes should follow the user
- Existing account login: server state should take priority, or user should be prompted

**Possible solutions:**
- On existing-account login, reset local state to server state
- Prompt the user to choose between local and server versions

src/components/Board/Board.actions.js
```diff
@@ -524,19 +531,259 @@ export function downloadImageFailure(message) {
...
+        // Graduate to SYNCED without pushing
+        dispatch(
+          updateBoard(
+            {
+              ...board,
+              syncStatus: SYNC_STATUS.SYNCED
+            },
+            true
+          )
+        );
```
RodriSanchez1, line 695: Maybe we should create a new reducer that dispatch only one action that updates all the boards. Instead of dispatching the `updateBoard` for each one.

Other option is mark the untracked local boards as `PENDING` and try to update all . But this will craete a lot of unnecesary requests to our api

@tomivm  what do you think

tomivm: looks good for the moment, we can improve it later to make it in a single req

src/components/Board/Board.actions.js
```diff
@@ -524,19 +531,259 @@ export function downloadImageFailure(message) {
...
+    for (const board of boardsToSync) {
+      try {
+        if (board.id.length < SHORT_ID_MAX_LENGTH) {
+          const newBoardId = await dispatch(
+            updateApiObjectsNoChild(board, true)
+          );
+          dispatch(replaceBoard(board, { ...board, id: newBoardId }));
+          if (activeBoardId === board.id) {
+            replaceHistoryWithActiveBoardId(getState);
+          }
```
tomivm, line 717: Maybe we can create a separate function for this part. For clarity only

src/components/Board/Board.actions.js
```diff
@@ -524,19 +531,259 @@ export function downloadImageFailure(message) {
...
+      await dispatch(
+        applyRemoteChangesToState({
+          boardsToAdd,
+          boardsToUpdate,
+          boardIdsToDelete
+        })
+      );
+
+      // 3. PUSH: Upload local changes + delete locally deleted boards from server
+      await dispatch(pushLocalChangesToApi(remoteBoards));
```
tomivm, line 778: I think this function can be splited in two separate functions one for push and another for delete locally deleted boards from server

PR #2101, opened 2026-02-06
Sync button component

magush27:
Created a new `SyncButton` component to the `CommunicatorToolbar`, providing improved visibility and control over board synchronization status. The changes also refactor how sync state is managed and displayed, replacing the previous spinner in the edit toolbar with a more informative sync button. Additionally, the sync status logic is enhanced to support multiple states and better offline handling.

**Sync functionality enhancements:**

* Added the new `SyncButton` component to `CommunicatorToolbar`, which visually communicates sync states (e.g., syncing, synced, pending, offline) and allows users to trigger sync actions.
* Introduced supporting files for `SyncButton`, including Redux container, constants, messages, and CSS for styling and internationalization.

**Sync status logic improvements:**

* Updated `Board.constants.js` to add a new `SYNCING` status, allowing for more granular sync state tracking.
* Refactored `CommunicatorToolbar` props and Redux mapping to support new sync-related properties, including `isLoggedIn` and improved state selection.

**UI and code cleanup:**

* Removed the old spinner from `EditToolbar.component.js` and unnecessary imports, as sync status is now handled by the new `SyncButton`.
* Cleaned up unused props in `CommunicatorToolbar.component.js` for clarity and maintainability.
* Imported colors from MUI that are not in the theme but are needed for the current sync displays.

**Sync state display improvements:**

* Now shows "Working Offline" with an OfflinePin icon when changes are made while offline
* Shows "Saved Locally" when back online but changes haven't been synced to the remote yet
* Once everything is synced, displays the green "Synced" cloud icon permanently
* Removed the manual sync button - synchronization now happens automatically


src/components/Communicator/CommunicatorToolbar/SyncButton/SyncButton.component.js
```diff
@@ -0,0 +1,179 @@
...
+      case DISPLAY_STATE.OFFLINE:
+        return intl.formatMessage(messages.offline);
+      case DISPLAY_STATE.WORKING_OFFLINE:
+        return intl.formatMessage(messages.workingOffline);
+      case DISPLAY_STATE.SAVING:
+        return intl.formatMessage(messages.saving);
+      case DISPLAY_STATE.SYNCED:
+        return intl.formatMessage(messages.synced);
+      case DISPLAY_STATE.PENDING:
+      case DISPLAY_STATE.SYNC:
```
RodriSanchez1: what about this case?

src/components/Communicator/CommunicatorToolbar/CommunicatorToolbar.component.js
```diff
@@ -219,7 +221,6 @@ class CommunicatorToolbar extends React.Component {
 CommunicatorToolbar.defaultProps = {
   className: '',
   boards: [],
-  isSelecting: false,
```
RodriSanchez1, line 222: this change is neccesary?

src/components/Communicator/CommunicatorToolbar/CommunicatorToolbar.container.js
```diff
@@ -81,10 +81,8 @@ class CommunicatorContainer extends React.Component {
   }
 }
 
-const mapStateToProps = (
-  { board, communicator, app: { userData, displaySettings } },
-  ownProps
-) => {
+const mapStateToProps = ({ board, communicator, app }, ownProps) => {
+  const { userData, displaySettings } = app;
   const activeCommunicatorId = communicator.activeCommunicatorId;
```
RodriSanchez1: this change is neccesary?

src/components/Communicator/CommunicatorToolbar/SyncButton/SyncButton.component.js
```diff
@@ -0,0 +1,184 @@
...
+  isSyncing,
+  isFetching,
+  hasPendingBoards,
+  onSyncClick
+}) => {
+  const [showSynced, setShowSynced] = useState(false);
+  const prevSyncStatusRef = useRef(null);
+
+  const getSyncStatus = () => {
+    if (isSyncing || isFetching) return SYNCING;
```
RodriSanchez1: It need to keep using the same `isSaving` state from the `EditToolbar.component.js` as it blocks the ui in differents way as the `isSyncing` or `isFetching` from here

RodriSanchez1 (changes requested):

RodriSanchez1:
Great progress on the SyncButton! After reviewing the display state logic, I found a couple of issues we should address:

1. "Working Offline" disappears on reconnect
When the user goes offline and makes changes, they see the "Working Offline" indicator. But as soon as the connection is reestablished, this state vanishes and falls through to the generic sync icon — the user has no idea they still have unsynced local changes. The "local changes pending" indicator should persist until the changes are actually synced to the cloud.

2. PENDING state is indistinguishable from the default SYNC state
When the user is online and has pending (unsynced) boards, getDisplayState() returns the sync status PENDING but visually it renders the same generic sync icon as the idle SYNC state. There's no visual distinction telling the user "you have local changes that need syncing." The DISPLAY_PENDING constant is defined in SyncButton.constants.js and even has a case in getAriaLabel(), but getDisplayState() never actually returns it — looks like this was intended but not fully wired up.

Suggestion: These two issues share the same root cause. We need a distinct "unsynced changes" visual state (different icon, color, or label) that works both offline and after reconnecting, and only clears once the sync actually completes successfully.

Let me know if you have any doubt

src/components/Communicator/CommunicatorToolbar/SyncButton/SyncButton.component.js
magush27: The Sync case is for manual syncing, it shows always except you are in another state

src/components/Communicator/CommunicatorToolbar/CommunicatorToolbar.component.js
magush27: No, I'm going to put it back in the next commits because the prop is still in use so it needs to be defined

src/components/Communicator/CommunicatorToolbar/CommunicatorToolbar.container.js
magush27: No, I reverted it to the original one to maintain the diff minimal in the next commits

src/components/Communicator/CommunicatorToolbar/SyncButton/SyncButton.component.js
magush27: Yes right, I'm checking the code and I see that `isSaving` is a local state in `Board.container.js` that tracks the individual board saves but `isSyncing` and `isFetching` are redux states. So I'm thinking that maybe is a good idea to move `isSaving` to redux but it would be a big refactor for `Board.container.js`. The other option is to pass `isSaving` as a prop though the components but it will mix redux data with prop drilling. What do you think?

magush27:
> Great progress on the SyncButton! After reviewing the display state logic, I found a couple of issues we should address:
> 
>     1. "Working Offline" disappears on reconnect
>        When the user goes offline and makes changes, they see the "Working Offline" indicator. But as soon as the connection is reestablished, this state vanishes and falls through to the generic sync icon — the user has no idea they still have unsynced local changes. The "local changes pending" indicator should persist until the changes are actually synced to the cloud.
> 
>     2. PENDING state is indistinguishable from the default SYNC state
>        When the user is online and has pending (unsynced) boards, getDisplayState() returns the sync status PENDING but visually it renders the same generic sync icon as the idle SYNC state. There's no visual distinction telling the user "you have local changes that need syncing." The DISPLAY_PENDING constant is defined in SyncButton.constants.js and even has a case in getAriaLabel(), but getDisplayState() never actually returns it — looks like this was intended but not fully wired up.
> 
> 
> Suggestion: These two issues share the same root cause. We need a distinct "unsynced changes" visual state (different icon, color, or label) that works both offline and after reconnecting, and only clears once the sync actually completes successfully.
> 
> Let me know if you have any doubt

I'm thinking that maybe we can put an icon like this one without the text to show the pending unsynced changes when being online again:
imagen

What do you think @RodriSanchez1 ?

src/components/Communicator/CommunicatorToolbar/SyncButton/SyncButton.component.js
RodriSanchez1: ok!

src/components/Communicator/CommunicatorToolbar/SyncButton/SyncButton.component.js
RodriSanchez1: we can pass it by props for now and create an issue to work on that later

RodriSanchez1:
Yes it can be that one! It should be showed always that the user have local changes. No matters if is online or not

magush27:
> Yes it can be that one! It should be showed always that the user have local changes. No matters if is online or not

The thing is that if the user is offline and has local changes it shows the working offline icon. If it's only offline it shows the offline icon. So we should replace the working offline icon with this one? I thought this one was only for when you have pending changes and are online

src/components/Communicator/CommunicatorToolbar/SyncButton/SyncButton.component.js
magush27: great! I'm solving this in next commits

RodriSanchez1:
Yes! I think that we should replace it! Copy the google docs funtionality behaviour

src/components/Communicator/CommunicatorToolbar/SyncButton/SyncButton.js
```diff
@@ -0,0 +1,187 @@
...
+  if (displayState === SAVED_LOCALLY) {
+    return (
+      <Button
+        className={baseClassName}
+        aria-label={getAriaLabel()}
+        disabled={isSyncDisabled}
+        onClick={onSyncClick}
+        style={{ color: amber[500] }}
+      >
+        <span className="SyncButton__label">
```
tomivm, line 115: The texts that uses SyncButton__label are not centered

tomivm: image

src/components/Communicator/CommunicatorToolbar/CommunicatorToolbar.css
```diff
@@ -55,6 +55,7 @@
 .CommunicatorToolbar__group--end {
   order: 2;
   justify-content: flex-end;
+  height: 100%;
 }
```
tomivm, line 59: Maybe this change is not necessary

PR #2108, opened 2026-02-13
Prepare AppContainer for visibility and connectivity changes to be used for Sync

tomivm:
Improve AppContainer to handle visibility changes and resume events, incorporating data refresh logic for logged-in users. Refactor data refresh handling to include debounce logic for synchronization during connectivity changes.

### DIFF
--- a/src/components/App/App.container.js
+++ b/src/components/App/App.container.js
@@ -114,10 +114,16 @@ export class AppContainer extends Component {
       const { updateConnectivity } = this.props;
 
       this.handleOffline = () => {
+        if (navigator.onLine) {
+          return;
+        }
         updateConnectivity({ isConnected: false });
       };
 
       this.handleOnline = () => {
+        if (!navigator.onLine) {
+          return;
+        }
         updateConnectivity({ isConnected: true });
         this.handleDataRefresh('Connection restored');
       };
--- a/src/components/App/App.container.js
+++ b/src/components/App/App.container.js
@@ -59,7 +59,7 @@ export class AppContainer extends Component {
 
   constructor(props) {
     super(props);
-    this.syncDebounceTimer = null;
+    this.lastSyncTime = null;
     this.handleOnline = null;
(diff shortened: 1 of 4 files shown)


src/components/App/App.container.js
```diff
@@ -94,33 +112,94 @@ export class AppContainer extends Component {
...
     configureConnectionStatus();
+
+    this.handleDataRefresh = (source = 'Unknown') => {
+      const { isLogged } = this.props;
+
+      if (!isLogged) {
+        return;
+      }
+
+      if (!window.navigator.onLine) {
```
RodriSanchez1: why dont you use the `isConnected` redux state?

src/components/App/App.container.js
```diff
@@ -94,33 +112,94 @@ export class AppContainer extends Component {
...
+      this.handleWebVisibilityChange
+    );
+  }
+
+  componentWillUnmount() {
+    if (this.syncDebounceTimer) {
+      clearTimeout(this.syncDebounceTimer);
+    }
+
+    cleanUpCvaOnResume(() => this.handleDataRefresh('App resumed'));
```
RodriSanchez1: You are calling a function that is defined on the `componentDidMount` , did test this PR? Not sure if you define a function using `this`, make the function available for al the component

src/components/App/App.container.js
tomivm: what is the difference?

src/components/App/App.container.js
tomivm: Is not a problem. But I find a bug, because i was creating different arrow function for pass as param to `onCvaResume` and other for `cleanUpCvaOnResume`. Now both are using the same function to correctly remove event listener

src/components/App/App.container.js
RodriSanchez1: there is no difference but we have dedicated state for that, why not using it?

PR #2109, opened 2026-02-13
Fix offline boards history

magush27:
Improves how board synchronization and navigation are handled in the application. The main changes include better error handling during board synchronization, ensuring the UI reflects the correct board state after API updates, and keeping navigation history consistent when board IDs change.

**Board synchronization and error handling:**

* Improved the `BoardContainer` component to handle errors gracefully when synchronizing boards with the API, and to ensure the correct board is selected if accessed via a URL (`src/components/Board/Board.container.js`).

**Navigation and state consistency:**

* Updated the `componentDidUpdate` lifecycle method in `BoardContainer` to synchronize the URL and UI state with the current board after API objects are fetched, and to update the `isFixedBoard` state accordingly.
* Enhanced the `boardReducer` to update `boards`, `activeBoardId`, and `navHistory` when a board's ID changes after being created via the API, ensuring navigation and state remain consistent.

### DIFF
--- a/src/components/Board/Board.actions.js
+++ b/src/components/Board/Board.actions.js
@@ -653,6 +653,7 @@ export function applyRemoteChangesToState({
  */
 export function pushLocalChangesToApi(remoteBoards = []) {
   return async (dispatch, getState) => {
+    const userEmail = getState().app?.userData?.email;
     const { boards, activeBoardId } = getState().board;
 
     // Boards explicitly marked PENDING by the sync system.
--- a/src/components/Board/Board.actions.js
+++ b/src/components/Board/Board.actions.js
@@ -651,7 +651,7 @@ export function applyRemoteChangesToState({
  */
 export function pushLocalChangesToApi() {
   return async (dispatch, getState) => {
-    const { boards } = getState().board;
+    const { boards, activeBoardId } = getState().board;
 
     // Separate boards to sync vs boards to delete
     const boardsToSync = boards.filter(
@@ -663,7 +663,13 @@ export function pushLocalChangesToApi() {
     for (const board of boardsToSync) {
       try {
         if (board.id.length < SHORT_ID_MAX_LENGTH) {
-          await dispatch(updateApiObjectsNoChild(board, true));
+          const newBoardId = await dispatch(
+            updateApiObjectsNoChild(board, true)
+          );
+          dispatch(replaceBoard(board, { ...board, id: newBoardId }));
+          if (activeBoardId === board.id) {
+            replaceHistoryWithActiveBoardId(getState);
+          }
         } else {
           await dispatch(updateApiBoard(board));
         }


src/components/Board/Board.container.js
```diff
@@ -229,11 +229,22 @@ export class BoardContainer extends Component {
...
 
-    // Loggedin user?
     if ('name' in userData && 'email' in userData && window.navigator.onLine) {
-      //synchronize communicator and boards with API
+      if (id) {
+        const urlBoard = boards.find(b => b.id === id);
+        if (urlBoard) {
+          changeBoard(id);
+        }
+      }
```
RodriSanchez1: This is not handlded by the code thats below?

src/components/Board/Board.container.js
```diff
@@ -326,8 +337,23 @@ export class BoardContainer extends Component {
...
+    } = this.props;
+    const { isGettingApiObjects } = this.state;
+
+    if (prevState.isGettingApiObjects && !isGettingApiObjects && board) {
+      if (id !== board.id) {
+        history.replace(`board/${board.id}`);
+      }
+      this.setState({ isFixedBoard: !!board.isFixed });
+    }
```
RodriSanchez1: why do we need to check the `isGettingApiObjects`? What if we just compare if the active board changed and the  url `id` is different than the active board id?

src/components/Board/Board.container.js
```diff
@@ -229,11 +229,22 @@ export class BoardContainer extends Component {
...
+
       this.setState({ isGettingApiObjects: true });
-      getApiObjects().then(() => this.setState({ isGettingApiObjects: false }));
+      try {
+        await getApiObjects();
+      } catch (err) {
+        console.error('Error syncing boards:', err);
+      }
+      this.setState({ isGettingApiObjects: false });
+      return;
```
RodriSanchez1: i think this modification are not really neccesary. What do you think?

src/components/Board/Board.container.js
```diff
@@ -326,8 +337,23 @@ export class BoardContainer extends Component {
...
+        params: { id }
+      }
+    } = this.props;
+    const { isGettingApiObjects } = this.state;
+
+    if (prevState.isGettingApiObjects && !isGettingApiObjects && board) {
+      if (id !== board.id) {
+        history.replace(`board/${board.id}`);
+      }
+      this.setState({ isFixedBoard: !!board.isFixed });
```
RodriSanchez1: this is really neccesary? can not be handled by the below if?

src/components/Board/Board.container.js
magush27: The code below is inside the if block that returns early for online users, so it never runs. But since the reducer is fixed to update `activeBoardId` in `CREATE_API_BOARD_SUCCESS` this might not be necessary anymore, will change it in the next commit

src/components/Board/Board.container.js
magush27: You're right, I'll implement this in the next commit

src/components/Board/Board.container.js
magush27: I think it's necessary because the original `.then()` caused a race condition because sync was happening in the background while the component continued.. maybe I can remove the try/catch since getApiObjects already has error handling

src/components/Board/Board.container.js
magush27: you're right, will simplify it

src/components/Board/Board.container.js
RodriSanchez1: Yes, I understand. We implemented this change before and this produced a double rerender of the board. So we needed to rollback it.  What prbolem we can face if the boards are synced in background? Our main issue is not resolved in the `componentDidUpdate`?

magush27: You're right, our main issue is resolved in `componentDidUpdate` and we shouldn't have more problems. I'll rollback it in the next commit

PR #2119, opened 2026-02-26
Solve sync problem about fields in boards

magush27:
Improvements to the board synchronization logic around handling core boards and boards created offline, as well as updates to the reducer and tests to ensure correct sync status and navigation. The most significant changes are grouped below.

**Synchronization and Board Transformation Logic:**

* Boards with a `PENDING` sync status that are either core boards (`support@cboard.io`) or have an empty email are now transformed to belong to the current user before being synced. This includes updating their `email`, `author`, and `name` fields. These boards are then pushed to the API as the user's own boards.

**Reducer and Sync Status Updates:**
* Any tile operation (create, delete, edit) now marks the board's `syncStatus` as `PENDING`, ensuring that changes are recognized and queued for synchronization.
* Corresponding updates in the reducer tests verify that the `syncStatus` is properly set to `PENDING` after tile operations and board creation.

### DIFF
--- a/src/components/Board/Board.actions.js
+++ b/src/components/Board/Board.actions.js
@@ -53,6 +53,7 @@ import API from '../../api';
 
 import {
   replaceBoardCommunicator,
+  addBoardCommunicator,
   upsertCommunicator,
   getApiMyCommunicators,
   editCommunicator,
@@ -648,21 +649,48 @@ export function applyRemoteChangesToState({
  * PUSH: Upload local changes to the API.
  * Pushes all boards with syncStatus: PENDING, plus untracked boards (no syncStatus)
  * that are newer than their remote version or don't exist on the server.
+ * - Core boards (support@cboard.io) with PENDING → transformed to user's boards, then synced
  * - Short ID boards (locally created) → updateApiObjectsNoChild (creates on server)
  * - Long ID boards (existing) → updateApiBoard (updates on server)
  */
 export function pushLocalChangesToApi(remoteBoards = []) {
   return async (dispatch, getState) => {
     const userEmail = getState().app?.userData?.email;
+    const userName = getState().app?.userData?.name || '';
     const { boards, activeBoardId } = getState().board;
 
     // Boards explicitly marked PENDING by the sync system.
-    // Only push boards that belong to the current user.
-    const pendingBoards = boards.filter(b => {
-      if (b.syncStatus !== SYNC_STATUS.PENDING || b.isDeleted) return false;
-      if (b.email !== userEmail) return false;
-      return true;
-    });
+    const pendingBoards = [];
+    for (const b of boards) {
+      if (b.syncStatus !== SYNC_STATUS.PENDING || b.isDeleted) continue;
+      if (!userEmail) continue;
+
+      if (!b.email || b.email === 'support@cboard.io') {
+        const extractName = () => {
+          if (b.name) return b.name;
+          if (b.nameKey) {
+            const splitedNameKey = b.nameKey.split('.');
+            return splitedNameKey[splitedNameKey.length - 1];
+          }
+          return 'Untitled Board';
+        };
+
+        const transformedBoard = {
+          ...b,
+          email: userEmail,
+          author: b.author || userName || userEmail,
+          name: extractName(),
+          isPublic: false
+        };
+        dispatch(updateBoard(transformedBoard, true));
+        pendingBoards.push(transformedBoard);
+        continue;
+      }
+      if (b.email === userEmail) {
+        pendingBoards.push(b);
+        continue;
+      }
+    }
 
     // Untracked boards (no syncStatus) that belong to the current user.
     const untrackedBoards = boards.filter(b => {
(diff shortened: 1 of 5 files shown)


src/components/Board/Board.actions.js
```diff
@@ -648,21 +649,48 @@ export function applyRemoteChangesToState({
...
+    const pendingBoards = [];
+    for (const b of boards) {
+      if (b.syncStatus !== SYNC_STATUS.PENDING || b.isDeleted) continue;
+      if (!userEmail) continue;
+
+      if (!b.email || b.email === 'support@cboard.io') {
+        const extractName = () => {
+          if (b.name) return b.name;
+          if (b.nameKey) {
+            const splitedNameKey = b.nameKey.split('.');
```
tomivm, line 672: namekey is the ke to be used with translations

src/components/Board/__tests__/Board.actions.test.js
```diff
@@ -907,26 +907,33 @@ describe('pushLocalChangesToApi', () => {
     expect(actionTypes).not.toContain(types.CREATE_API_BOARD_STARTED);
   });
 
-  it('should not push PENDING default boards (support@cboard.io)', async () => {
-    const defaultBoard = {
+  it('should transform and push PENDING core boards (support@cboard.io)', async () => {
+    const coreBoard = {
```
tomivm: `defaultBoard` is the correct name, core boards are AAC boards with flexible words to say a lot of things with less words

magush27: Yes, you're right! I corrected it in the next commit

src/components/Board/Board.actions.js
magush27: Yes, here we are using it to extract the board name from the nameKey in case it doesn't have a name. It takes the last part of the nameKey after splitting it by dots and uses it directly as the board name when the boards doesn't have a name

src/components/Board/Board.actions.js
```diff
@@ -648,21 +649,48 @@ export function applyRemoteChangesToState({
  * PUSH: Upload local changes to the API.
  * Pushes all boards with syncStatus: PENDING, plus untracked boards (no syncStatus)
  * that are newer than their remote version or don't exist on the server.
+ * - Core boards (support@cboard.io) with PENDING → transformed to user's boards, then synced
```
tomivm, line 652: ` * - Default boards (support@cboard.io) with PENDING → transformed to user's boards, then synced`

src/components/Board/Board.actions.js
```diff
@@ -648,21 +649,48 @@ export function applyRemoteChangesToState({
...
-    // Only push boards that belong to the current user.
-    const pendingBoards = boards.filter(b => {
-      if (b.syncStatus !== SYNC_STATUS.PENDING || b.isDeleted) return false;
-      if (b.email !== userEmail) return false;
-      return true;
-    });
+    const pendingBoards = [];
+    for (const b of boards) {
+      if (b.syncStatus !== SYNC_STATUS.PENDING || b.isDeleted) continue;
+      if (!userEmail) continue;
```
tomivm, line 666: this comprobation could be before the `for` loop

src/components/Board/Board.actions.js
tomivm: I mean, the last part of the nameKey is the name in english, in order to set the name in the language of the user, we can use the translation function with the nameKey. 
Now we are not having a way to use `intl` on actions. 
We can continue like this or maybe pass intl as param for actions ( No Strictly necessary now)

src/components/Board/Board.actions.js
```diff
@@ -648,21 +649,48 @@ export function applyRemoteChangesToState({
...
 
     // Boards explicitly marked PENDING by the sync system.
-    // Only push boards that belong to the current user.
-    const pendingBoards = boards.filter(b => {
-      if (b.syncStatus !== SYNC_STATUS.PENDING || b.isDeleted) return false;
-      if (b.email !== userEmail) return false;
-      return true;
-    });
+    const pendingBoards = [];
+    for (const b of boards) {
```
tomivm, line 664: Instead of create a new for loop to filter `b.syncStatus !== SYNC_STATUS.PENDING || b.isDeleted` and update board properties. We can leave the filter function like the old version and loop the filtered `pendingBoards` with the new logic for `if (!b.email || b.email === 'support@cboard.io')` case.
I think that is more clear and we have a better separation of concerns

src/components/Board/Board.actions.js
```diff
@@ -708,10 +736,9 @@ export function pushLocalChangesToApi(remoteBoards = []) {
     for (const board of boardsToSync) {
       try {
         if (board.id.length < SHORT_ID_MAX_LENGTH) {
-          const newBoardId = await dispatch(
-            updateApiObjectsNoChild(board, true)
-          );
-          dispatch(replaceBoard(board, { ...board, id: newBoardId }));
+          // Add board to communicator first so replaceBoardCommunicator can find it
+          dispatch(addBoardCommunicator(board.id));
+          await dispatch(updateApiObjectsNoChild(board, true));
```
tomivm, line 741: this line is not necessary now?
 `dispatch(replaceBoard(board, { ...board, id: newBoardId }));`

src/components/Board/Board.actions.js
magush27: Sure, I thought about the `intl` thing that doesn't work in actions now. I think maybe it's ok to leave it like it's now and later check other options like pass `intl` as a param for actions

src/components/Board/Board.actions.js
magush27: Yes, solved this in the next commits

src/components/Board/Board.actions.js
magush27: right! solved this in the next commits

src/components/Board/Board.actions.js
magush27: Done! I refactored it to use the filter function for getting the pending boards first, then a separate loop to transform the default boards. This way the filtering logic and the transformation logic are clearly separated.

src/components/Board/Board.actions.js
magush27: Right, deleted it in next commits

PR #2120, opened 2026-03-02
Cleaner arch sync

tomivm:
Separate the state of syncronization from boards propperties.
Also create a middleware to mark boards as outdated


src/components/Board/Board.reducer.js
```diff
@@ -158,7 +179,7 @@ function boardReducer(state = initialState, action) {
         activeBoardId: action.boardId,
         isFixed: fixed
       };
-    case UPDATE_BOARD:
+    case UPDATE_BOARD: {
```
RodriSanchez1, line 176: not sure if it the reducer need the `{}` wrap

src/components/Board/Board.actions.js
```diff
@@ -654,19 +631,20 @@ export function applyRemoteChangesToState({
 export function pushLocalChangesToApi(remoteBoards = []) {
   return async (dispatch, getState) => {
     const userEmail = getState().app?.userData?.email;
-    const { boards, activeBoardId } = getState().board;
+    const { boards, activeBoardId, syncMeta } = getState().board;
```
magush27: I think you should add this to avoid issues when when no user is logged in
```suggestion
if (!userEmail) return;
```

src/components/Board/Board.actions.js
```diff
@@ -551,52 +557,6 @@ export function getApiMyBoards() {
...
 }
 
-/**
- * Classify remote boards for PULL operation.
- * Identifies boards that are new from the server or have newer versions on the server.
- * @param {Array} localBoards - Boards from local state
- * @param {Array} remoteBoards - Boards from the server
- * @returns {{ boardsToAdd, boardsToUpdate, boardIdsToDelete }}
- */
-export function classifyRemoteBoards(localBoards, remoteBoards) {
```
RodriSanchez1, line 561: Is there a reason to move the function to the `Board.utils.js`? If not we can leave it here within all the related function definitions or maybe create a specific file for the sync. Something like `Board.actions.synchronization.js`

src/components/Board/Board.reducer.js
```diff
@@ -254,32 +283,39 @@ function boardReducer(state = initialState, action) {
...
-      });
+      const newBoard = { ...action.boardData, lastEdited: moment().format() };
+      nextBoards.push(newBoard);
       return {
         ...state,
-        boards: nextBoards
+        boards: nextBoards,
+        syncMeta: setSyncMeta(state.syncMeta, action.boardData.id, {
+          status: SYNC_STATUS.PENDING,
+          isDeleted: false
```
RodriSanchez1: The `isDeleted` prop is not being sent to `setSyncMeta` consistently. Consider only sending it during `DELETE_BOARD` operations.

RodriSanchez1:
## Test Review: `*.mapState.test.js` files

The PR adds 6 new `*.mapState.test.js` files that test `mapStateToProps` across different containers. After reviewing them, **most are redundant** — they all test the same thing: that `getVisibleBoards()` filters out soft-deleted boards.

### What's already covered

`Board.selectors.test.js` already thoroughly tests `getVisibleBoards` with 4 cases:
- All boards visible when none deleted
- Filters out `isDeleted: true`
- Includes untracked boards (no syncMeta entry)
- Returns empty when all deleted

Since every container just calls `getVisibleBoards()`, testing the same filter behavior 6 more times doesn't add confidence — it adds maintenance burden.

### Recommendation

**Keep** the selector tests in `Board.selectors.test.js` — they're thorough and this is the correct place to test this behavior.

**Move** the Board.container and CommunicatorDialog.container mapState tests into their respective existing test files rather than creating separate `*.mapState.test.js` files:
- `Board.container.mapState.test.js` has a useful extra test (active board returns `undefined` when deleted) — move this into the existing Board container test file
- `CommunicatorDialog.container.mapState.test.js` tests `communicatorBoards` filtering, which involves extra logic beyond the selector — move this into the existing CommunicatorDialog container test file

**Remove** the other 4 files — they test nothing beyond what the selector test already covers:
- `Analytics.container.mapState.test.js`
- `Export.container.mapState.test.js`
- `Import.container.mapState.test.js`
- `CommunicatorToolbar.container.mapState.test.js`

This reduces 7 test files for the same behavior down to 1 selector test + 2 container-specific tests co-located with their existing tests.

RodriSanchez1 (commented):
Good work! I left some comments,!

RodriSanchez1:
We should consider removing the `IMPORT_BOARD` reducer as it is not being used and can cause issues if is used without taking care  the new `metaSync` state

src/components/Board/Board.actions.js
tomivm: 1. Pure Utility Function
   - No side effects
   - Deterministic: same inputs → same outputs
   - Easy to test in isolation (as shown by the 14 test cases)
2. Reusability
   - Currently used in Board.actions.js (sync logic)
   - Tested extensively in Board.actions.test.js
   - Could potentially be used in selectors or other modules
3. Separation of Concerns
   - Keeps complex classification logic out of the action creator
   - Makes the action creator more readable (see line 736: clean destructuring)
   - Business logic is isolated from Redux machinery
4. Testability
   - Can be tested independently without mocking Redux
   - The 14 test cases verify edge cases without needing a Redux store
   - Makes CI/CD faster and tests more focused
5. Follows Module Pattern
   - Consistent with existing utils like isLocalBoard and isServerBoard
   - Part of a cohesive sync utility family
   - Aligns with the Board module's architecture (.actions, .reducer, .selectors, .utils)
📊 Current Usage Pattern:
Board.utils.js (definition)
    ↓
Board.actions.js (imports & uses in syncBoards action)
    ↓
Board.actions.test.js (unit tests)
🎯 What Would Happen If You Moved It OUT?
If you moved it into Board.actions.js:
- ❌ Harder to test (would need to mock Redux)
- ❌ Couples business logic with action creators
- ❌ Violates single responsibility principle
- ❌ Makes Board.actions.js even larger (already 1119 lines!)
💡 Conclusion:
Keep it in Board.utils.js. The function is:
- A pure, testable utility
- Well-documented with clear purpose
- Properly separated from side effects
- Following best practices for Redux architecture

src/components/Board/Board.reducer.js
tomivm: The UPDATE_BOARD case (starting at line 182) does have curly braces {}, and this is correct and necessary because:
1. Block scope is needed: The case declares block-scoped variables (const updateBoards, const oldBoard, const index, const nextBoard)
2. Prevents variable hoisting conflicts: Without braces, these const declarations would be in the same scope as other cases, potentially causing naming conflicts
3. Best practice: When a case has multiple statements or variable declarations, it should use a block scope

src/components/Board/Board.actions.js
tomivm: Added 🚀

src/components/Board/Board.container.js
```diff
@@ -1750,8 +1751,8 @@ const mapStateToProps = state => {
   const offlineVoiceAlert = !isConnected && speech.options.isCloud;
   return {
     communicator: currentCommunicator,
-    board: board.boards.find(board => board.id === activeBoardId),
-    boards: board.boards,
+    board: getVisibleBoards(state).find(board => board.id === activeBoardId),
+    boards: getVisibleBoards(state),
```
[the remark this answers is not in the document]
tomivm: I cannot reproduce the bug. There is a possibility that during a board container didUpdate this occurs. We can plan a propper solution for avoid infinite spinner.
@RodriSanchez1 can you reproduce it ?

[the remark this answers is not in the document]
RodriSanchez1: This cannot occur now because is not possible to delete an active board. The `CommunicatorDialogBoardItem.component.js` has the following condition on teh Delete Icon render. That's the only possible way to delete a board.

```
<IconButton
    disabled={
        communicator.rootBoard === board.id ||
        activeBoardId === board.id
    }
```

RodriSanchez1:
@tomivm please remember to check the above [comment](https://github.com/cboard-org/cboard/pull/2120#issuecomment-3987473388)

also I was thinking that we should add a redux state migration to add the new `syncMeta`  prop as we don on the `reducer.js`

https://github.com/cboard-org/cboard/blob/7b8969fe81f3198be695d9f1717ba755e574a676/src/reducers.js#L114-L132 

 What do you think?

PR #2121, opened 2026-03-03
Add skill for normalizing Redux state

tomivm:
Add skill to normalize redux state


martinbedouret (changes requested):
Please review @tomivm

PR #2122, opened 2026-03-03
Fix sync for default edited boards and offline created boards

magush27:
Fixes synchronization for boards that were modified while logged out or boards based on default boards with emal support@cboard.io.

**Key changes:**
- Default boards (support@cboard.io) modified by the user are now transformed to belong to the current user and synced to the server as duplicated
- Boards created offline are transformed to belong to the user and synced when the user logs in
- Transformed boards are CREATED on the server (not updated) since they don't exist for the user yet
- Added `addBoardCommunicator` before creating boards to ensure they appear in the communicator
- Unmodified default boards (no syncMeta entry) are no longer unnecessarily synced


src/components/Board/Board.actions.js
```diff
@@ -625,41 +626,89 @@ export function applyRemoteChangesToState({
...
+
+    // Helper to transform default/offline boards to belong to the current user
+    const transformBoard = board => {
+      const transformedBoard = {
+        ...board,
+        email: userEmail,
+        author: board.author || userName || userEmail,
+        name: extractName(board),
+        isPublic: false
+      };
```
RodriSanchez1: we could extract this functionality from the `updateApiMarkedBoards`  and avoid code repetition?

magush27: RIght! I've extracted it and put it into `Board.utils.js`

RodriSanchez1: I just checked! You can also extract the boards transformation, where the `email`, `author` and `name` repetaed in both places. 
The  `locale` needs to be added here also. 

Not sure if the  `hidden`

magush27: Yes, in the next commits I've done that

src/components/Board/Board.actions.js
```diff
@@ -625,41 +627,79 @@ export function applyRemoteChangesToState({
  * PUSH: Upload local changes to the API.
  * Pushes all boards with syncStatus: PENDING, plus untracked boards (no syncStatus)
  * that are newer than their remote version or don't exist on the server.
+ * - Default boards (support@cboard.io) → transformed to user's boards, then CREATED on server
+ * - Offline boards (empty email) → transformed to user's boards, then CREATED on server
```
RodriSanchez1: The "offline" created boards have the `support@cboard.io` email as default. Maybe this comment should be:
"Unlogged boards within `support@cboard.io` as default → transformed to user's boards, then CREATED on server"

src/components/Board/Board.actions.js
```diff
@@ -625,41 +627,79 @@ export function applyRemoteChangesToState({
...
+
+    // Helper to transform default/offline boards to belong to the current user
+    const transformBoard = board => {
+      const transformedBoard = transformBoardForUser(
+        board,
+        userEmail,
+        userName,
+        locale
+      );
+      dispatch(updateBoard(transformedBoard, true));
```
RodriSanchez1: I think the `FROM_REMOTE` param needs to be `false`. If not the new stored board get the syncMeta state as `SYNCED` and not pushed to the api

src/components/Board/Board.actions.js
```diff
@@ -625,41 +627,79 @@ export function applyRemoteChangesToState({
...
+        continue;
+      }
+      if (b.email === userEmail) {
+        pendingBoards.push(b);
+        continue;
+      }
+    }
+
+    // Untracked boards (no syncMeta entry) that belong to the current user
+    // or were created offline (empty email).
```
RodriSanchez1: the same than above, the comment should refer to created unlogged boards

src/components/Board/Board.actions.js
```diff
@@ -625,41 +627,79 @@ export function applyRemoteChangesToState({
...
     const untrackedBoardsToSync = [];
     const remoteBoardMap = new Map(remoteBoards.map(b => [b.id, b]));
 
+    // Helper to transform board if needed
+    const transformBoardIfNeeded = board => {
+      if (!board.email || board.email === 'support@cboard.io') {
+        return transformBoard(board);
+      }
+      return board;
+    };
```
RodriSanchez1: I noticed that the same condition `if (!board.email || board.email === 'support@cboard.io')` is verified in the `pendingBoards`. Is there a way to use the same implementation in the `untrackedBoards`. In that way, we are going to use the same pattern in both places

PR #2159, opened 2026-04-21
Cboard access frontend

### LINKED ISSUE #2137: Cboard Access - Frontend Implementation
Frontend implementation for Cboard Access feature. This enables users to access business or institutions AAC boards via QR codes or from a listing within the app communicator menu.

Overview:
 - `AccessViewer` component for viewing boards in read-only mode
 - Instant internal navigation (all boards loaded in single request)
 - Company branding support (custom colors)
 - `CboardAccessList` for browsing available businesses
 - Integration with existing `SpeechProvider`
 - Deep linking support for mobile apps

RodriSanchez1:
Closes #2137


src/components/Board/Board.component.js
```diff
@@ -490,7 +503,7 @@ export class Board extends Component {
                       board.grid ? board.grid.columns : DEFAULT_COLUMNS_NUMBER
                     }
                     rows={board.grid ? board.grid.rows : DEFAULT_ROWS_NUMBER}
-                    dragAndDropEnabled={isSelecting}
+                    dragAndDropEnabled={!isAccessViewerMode && isSelecting}
                     renderItem={this.renderTileFixedBoard}
                     onItemDrop={onTileDrop}
                     fixedRef={this.fixedBoardContainerRef}
```
tomivm: Probably is better to replace the changes in this files for something like this 
```
render() {
  const { viewerMode, ...rest } = this.props;
  
  if (viewerMode) {
    return this.renderViewerMode();  // Simplified render path
  }
  
  return this.renderFullMode();  // Current render logic, extracted to method
}
renderViewerMode() {
  // ONLY: div.Board > div.Board__tiles > Grid/FixedGrid + ScrollButtons
  // 8 props needed vs 50+
}
```

tomivm (commented):
Looks Good!!

src/cordova-util.js
```diff
@@ -35,9 +35,35 @@ export const cleanUpCvaOnResume = onResume => {
...
+export const initDeepLinking = () => {
+  if (!window.IonicDeeplink) {
+    console.log('IonicDeeplink plugin not available');
+    return;
+  }
+
+  const history = require('./history').default;
+
+  window.IonicDeeplink.route(
+    { '/access/:slug/:code': { target: 'access' } },
```
tomivm: initDeepLinking() only registers ONE route:
{ '/access/:slug/:code': { target: 'access' } }
But apple-app-site-association tells iOS to intercept both /access/* and /board/*. So when a user with the app installed taps a /board/abc123 link, iOS opens the native app — but the app has no handler for that pattern → IonicDeeplink fires the nomatch callback → just logs to console → user lands on whatever screen was last open.

I think we can Fix in
two files, minimal changes:
1. src/cordova-util.js — add the /board/:id route
window.IonicDeeplink.route(
  {
    '/access/:slug/:code': { target: 'access' },
    '/board/:id': { target: 'board' }          // ← add this
  },
  match => {
    if (match.$route === '/access/:slug/:code') {
      history.push(`/access/${match.$args.slug}/${match.$args.code}`);
    }
    if (match.$route === '/board/:id') {        // ← add this
      history.push(`/board/${match.$args.id}`);
    }
  },
  nomatch => {
    console.log('Deep link no match:', nomatch.$link);
  }
);
This works because Board.container.js already listens to route param changes via UNSAFE_componentWillReceiveProps (line 284) — a history.push('/board/:id') is all it needs to load the right board.
2. public/.well-known/apple-app-site-association — keep /board/* but it's now legitimate
"paths": ["/access/*", "/board/*"]
Once the handler exists, this entry is intentional and correct. You can keep it.

public/.well-known/apple-app-site-association
```diff
@@ -0,0 +1,11 @@
+{
+  "applinks": {
+    "apps": [],
+    "details": [
+      {
+        "appID": "TEAM_ID.com.unicef.cboard",
```
tomivm: We should remember to replace with real one

src/cordova-util.js
```diff
@@ -35,9 +35,35 @@ export const cleanUpCvaOnResume = onResume => {
   document.removeEventListener('resume', onResume, false);
 };
 
+export const initDeepLinking = () => {
+  if (!window.IonicDeeplink) {
+    console.log('IonicDeeplink plugin not available');
+    return;
+  }
+
+  const history = require('./history').default;
```
tomivm: Here we have a Circular dependency. See this and tell me your thoughts: history.js imports isCordova from cordova-util.js, so cordova-util.js can't do a top-level import history from './history' — circular. The dynamic require('./history').default is the workaround.
Second issue: look at index.js line 44-47:
const renderApp = () => {
  if (isCordova()) {
    initCordovaPlugins();   // ← called BEFORE ReactDOM.render()
  }
  ReactDOM.render(...);     // ← React hasn't mounted yet
};
initDeepLinking registers a listener before React has rendered. It works because history is a singleton created at module load time and React Router reads from it on mount — but it's still a fragile ordering assumption.

The Proper Solution — Dependency Injection
The root cause of both problems is that cordova-util.js is reaching out to grab history itself. That's the wrong direction. The caller should provide it.
Step 1 — Refactor initDeepLinking to accept history as a parameter

export const initDeepLinking = (history) => {
  if (!window.IonicDeeplink) {
    console.log('IonicDeeplink plugin not available');
    return;
  }
  window.IonicDeeplink.route(
    { '/access/:slug/:code': { target: 'access' } },
    match => {
      if (match.$route === '/access/:slug/:code') {
        history.push(`/access/${match.$args.slug}/${match.$args.code}`);
      }
    },
    nomatch => {
      console.log('Deep link no match:', nomatch.$link);
    }
  );
};
No more require inside. No circular dep. The function signature is honest about what it needs.
Step 2 — Remove initDeepLinking from initCordovaPlugins
export const initCordovaPlugins = () => {
  console.log('now cordova is ready ');
  if (isCordova()) {
    // NO initDeepLinking here — that's a routing concern, not a plugin concern
    try { window.StatusBar.hide(); } catch (err) { ... }
    try { window.AndroidFullScreen.immersiveMode(...); } catch (err) { ... }
    try { configFacebookPlugin(); } catch (err) { ... }
    try { ... configAppPurchasePlugin / IOSconfigureInAppPurchasePlugin } catch (err) { ... }
  }
};
Step 3 — Wire it in index.js
import history from './history';
import { isCordova, onCordovaReady, initCordovaPlugins, initDeepLinking } from './cordova-util';
const renderApp = () => {
  if (isCordova()) {
    initCordovaPlugins();
    initDeepLinking(history);  // ← explicit, history is already a resolved singleton
  }
  ReactDOM.render(...);
};

src/components/Board/Board.utils.js
```diff
@@ -1,6 +1,188 @@
...
+  }
+
+  const newOutput = isLiveMode
+    ? [...output, tile, createLiveTile(generateId)]
+    : [...output, tile];
+
+  if (changeOutput) changeOutput(newOutput);
+
+  return { navigated: false };
+};
```
tomivm, line 146: processTileClick reorders vocalizeTile vs history.push
  On master, the order for folder navigation was changeBoard → history.push → say(). 
  New code does, changeBoard → vocalizeTile → onNavigate(history.push). Both are synchronous and
   the swap is unlikely to be user-visible, but it IS a behavior change ?

src/components/Board/Board.component.js
```diff
@@ -2,7 +2,6 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import keycode from 'keycode';
```
tomivm: This is never used

tomivm (commented):
Good job @magush27 !!!

src/components/Board/Board.component.js
magush27: This is already superseded because now Board.component.js now renders <BoardGrid /> and has zero awarness of viewer mode.

src/components/Board/Board.component.js
magush27: Solved this in next commits

src/components/Board/Board.utils.js
magush27: It's not a behaviour change. Solved this in next commits to match master's order.

src/cordova-util.js
magush27: Solved in next commits

src/cordova-util.js
magush27: Solved in next commits

public/.well-known/apple-app-site-association
magush27: Yes! I don't know it hehe

src/components/AccessViewer/AccessViewer.js
```diff
@@ -0,0 +1,278 @@
...
+      history.push(isLogged ? '/' : '/login-signup');
+    },
+    [isLogged, history, cancelSpeech]
+  );
+
+  if (loading) {
+    return (
+      <div className="AccessViewer__loading">
+        <CircularProgress size={60} thickness={5} />
+        <p>Loading board...</p>
```
RodriSanchez1: Add the translation!

src/components/AccessViewer/AccessViewerError.js
```diff
@@ -0,0 +1,44 @@
...
+    Icon: BlockIcon,
+    title: 'Access not allowed',
+    description: 'This board is not available with the provided code.'
+  },
+  error: {
+    Icon: ErrorIcon,
+    title: 'Connection error',
+    description:
+      'We could not load the board. Please check your connection and try again.'
+  }
```
RodriSanchez1, line 24: Add the translations for all this messages

src/components/AccessViewer/AccessViewerNavbar.js
```diff
@@ -0,0 +1,84 @@
...
+}) => {
+  return (
+    <div className="Navbar">
+      <div className="Navbar__group Navbar__group--start">
+        <BackButton disabled={disableBackButton} onClick={onBackClick} />
+        <IconButton
+          disabled={disableBackButton}
+          onClick={onHomeClick}
+          aria-label="home"
+          style={{ color: '#fff' }}
```
RodriSanchez1: Avoid applying this color to the icon, as it makes it difficult to distinguish when the icon is disabled.

src/components/AccessViewer/AccessViewer.css
```diff
@@ -0,0 +1,76 @@
...
+.AccessViewer__header-name {
+  font-size: 16px;
+  font-weight: 500;
+}
+
+.AccessViewer__output {
+  height: 112px;
+  max-height: 15%;
+  min-height: 70px;
+  border-bottom: 1px solid rgba(0, 0, 0, 0.12);
```
RodriSanchez1: this boarder is not neccesary. It add a white line between the `SymbolOutput` and the `Navbar` that is perceptible in the dark mode

src/components/Board/Output/SymbolOutput/SymbolOutput.js
```diff
@@ -47,11 +47,17 @@ class SymbolOutput extends PureComponent {
          */
         label: PropTypes.oneOfType([PropTypes.string, PropTypes.node])
       })
-    )
+    ),
+    /**
+     * If true, hides editing controls (remove, backspace, clear, share, live mode)
+     * and disables symbol writing. Used by AccessViewer.
+     */
+    isAccessViewerMode: PropTypes.bool
```
RodriSanchez1: The `Output.container` is not being passing this prop to the component. Should we avoid the changes on this files?

src/components/AccessViewer/AccessViewerNavbar.js
```diff
@@ -0,0 +1,84 @@
...
+import CloseIcon from '@material-ui/icons/Close';
+import IconButton from '@material-ui/core/IconButton';
+
+import BackButton from '../UI/BackButton';
+import LockToggle from '../UI/LockToggle';
+import SettingsButton from '../UI/SettingsButton';
+import '../Board/Navbar/Navbar.css';
+
+const AccessViewerNavbar = ({
+  intl,
```
RodriSanchez1: this prop is never read

src/cordova-util.js
```diff
@@ -35,6 +35,31 @@ export const cleanUpCvaOnResume = onResume => {
...
+      }
+      if (match.$route === '/board/:id') {
+        history.push(`/board/${match.$args.id}`);
+      }
+    },
+    nomatch => {
+      console.log('Deep link no match:', nomatch.$link);
+    }
+  );
+};
```
RodriSanchez1, line 60: We currently do not support the `IonicDeeplink` plugin in the Cboard repository. The plugin is already installed (`ionic-plugin-deeplinks`), but it has never actually been used in the Cboard App. Unfortunately, the plugin is now publicly archived, so we should consider removing it and replacing it with the plugin you suggested.

Could you make those changes in the Ccboard repository?

src/setupTests.js
```diff
@@ -3,6 +3,23 @@ import Adapter from '@wojtekmaj/enzyme-adapter-react-17';
 import { Stream } from 'stream';
 import 'fake-indexeddb/auto';
 
+if (typeof MessageChannel === 'undefined') {
```
RodriSanchez1: Do you get to the root cause of why the test failed and we needed to add this changes?

RodriSanchez1 (commented):
@magush27  Good Work! I left some comments. We need to remember to update hte `apple-app-site-association` file and ~~remove the changes on the `config.yml`~~

I think that we need to implement some test for this new implementation. What do you think?

magush27:
@RodriSanchez1 yes! I think it's a good idea to add tests regarding this implementation, I'll do that after solving the other comments you left

src/components/AccessViewer/AccessViewer.js
magush27: yes! solved in next commits

src/components/AccessViewer/AccessViewerError.js
magush27: yes! solved in next commits

src/components/AccessViewer/AccessViewerNavbar.js
magush27: great, solved in next commits!

src/components/AccessViewer/AccessViewerNavbar.js
magush27: solved!

src/components/AccessViewer/AccessViewer.css
magush27: solved!

src/components/Board/Output/SymbolOutput/SymbolOutput.js
magush27: It was hiding some controls like share, backspace and clear. But I think in the new architecture this is not needed anymore. I will revert this part as navigationSettings are already hiding and showing those controls

src/setupTests.js
magush27: Yes, the MessageChannel mock was added because the ogv library (version ^1.8.9, a WebAssembly audio/video decoder) uses MessageChannel internally. Jest runs in a Node.js/jsdom environment which doesn't implement MessageChannel, so any test that indirectly imports ogv would crash with MessageChannel is not defined. This prevents the crash without actually implementing the full API

src/cordova-util.js
magush27: I don't remember mentioning a replacing plugin, but I think it's a good idea to handle this natively instead of using a dependency. I'm thinking to handle it this way:

- On iOS: implement `handleOpenURL` (built into Cordova for custom URL schemes) + `apple-app-site-association` for Universal Links
- On Android: handle the intent via `window.handleOpenURL` or the `resume` event

What do you think?

src/setupTests.js
RodriSanchez1: I find it a bit strange that the tests didn’t fail when the OGV import was still inside `Board.container`.

src/cordova-util.js
RodriSanchez1: Not sure if this can be implemented without a Cordova plugin. If that’s an option, go with it.

src/setupTests.js
magush27: I found a more accurate root cause:
The tests dind't fail before becase `Board.container.test.js` has `jest.mock('ogv', ...)` at line 4 and it mocks the OGV module so the real one (which uses MessageChannel) never loads. That's why those tests were always fine.                                                                                                                                                                          
                                                                                                                                                                                                  They started failing now because OGV was moved from `Board.container.js` into `Board.utils.js` in commit 5a8535e8 ("Import pure utility functionalities"). Board.utils.js is now imported by more things across the codebase. Any test that transitively imports `Board.utils.js` without the ogv mock will load the real OGV library, which hits MessageChannel and crashes in jsdom.                     
                                                                                                                                                                                                   So the `setupTests.js` polyfill it's a global solution that covers all current and future test files that pull in `Board.utils.js`. The alternative would be adding `jest.mock('ogv', ...)` to every test file that imports `Board.utils.js`, which is much more fragile.

PR #2193, opened 2026-05-20
Manage how to sync unauthenticated changes

magush27:
Warn unauthenticated users on edit and discard local changes on login for existing users that already have boards saved remotely.                                                                                                      

**Changes:**
- Show a warning dialog when an unauthenticated user tries to enter edit mode or add a tile, informing them that changes won't be saved and offering options to log in/sign up or continue without saving. 
- Reused `LoginRequiredModal` component with new optional props (open, onClose, onContinue, title, text) to support this controlled usage without breaking existing behavior.
- Added 3 new i18n messages for the warning dialog (unauthenticatedEditTitle, unauthenticatedEditText, continueWithoutSaving) to `cboard.json`.
- On login, discard any local unauthenticated changes (boards and communicators) if the user is not a first-time user or already has saved communicators on the backend, restoring their remote state instead.


src/components/Account/Login/Login.actions.js
```diff
@@ -176,7 +176,12 @@ export function login({ email, password, activatedData }, type = 'local') {
         currentCommunicator =
           loginData.communicators[lastRemoteSavedCommunicatorIndex]; //use the latest communicator
       }
-      if (apiBoards.length && currentCommunicator) {
+      const hasRemoteCommunicators =
+        loginData.communicators && loginData.communicators.length > 0;
+      const isFirstLogin = loginData.isFirstLogin === true;
+      const discardLocalChanges = !isFirstLogin || hasRemoteCommunicators;
+
+      if (apiBoards.length && currentCommunicator && !discardLocalChanges) {
```
RodriSanchez1: If the user has any apiBoard should be added to the local state. Remove the `!discardLocalChanges` from the comprobation

magush27: The remote boards are actually added to local state in the discard path just through a different route that is `loginSuccess` that spreads the full `loginData` intro it's payload, which includes `loginData.boards`. The reducer `LOGIN_SUCCESS` discard branch then picks them up via `action.payload.boards` and builds the new board state from them. If we remove `!discardLocalChanges` then `addBoards(apiBoards)` would run first and make `loginSuccess` to inmediately overwrite `state.boards` entire in the discard branch, so the `ADD_BOARDS` dispatch would have a redundant render. The guard keeps the two login flows cleanly separated, in one side the non-discard path loads boards via `ADD_BOARDS` and discard path is self-contained in the `LOGIN_SUCCESS` reducer case.. What do you think?

RodriSanchez1:
## Why the `discardActiveBoardId` fallback is needed

When a user logs in and chooses to **discard their local changes**, the reducer replaces the entire board state with the remote boards plus the built-in defaults. At that point, `activeBoardId` still holds whatever board the user was viewing *before* login — which may have been a locally-created board that was just wiped out.

If we blindly keep that stale `activeBoardId`, the app ends up pointing at a board ID that no longer exists in state, causing a blank/broken board view. The most common scenario is a **first-time login with no remote communicators**: the user was browsing the default local board set, then logs in and the discard path removes local boards that don't exist on the server.

The fix checks whether `activeBoardId` is still present in the new `allBoards` set and, if not, falls back to `'root'` — the default root board that is always guaranteed to be present after a discard (it comes from `initialBoardsState`).

```js
const allIds = new Set(allBoards.map(b => b.id));
const discardActiveBoardId = allIds.has(activeBoardId)
  ? activeBoardId
  : 'root';
```

Without this guard, `navHistory` would be seeded with a gone board ID, and the next render would crash or show nothing.

Issue #2207, opened 2026-06-01
refactor(sync): per-domain sync state machine for communicators (mirror board syncMeta)


tomivm:
## Problem

The sync state (`syncMeta`: status pending/syncing/synced + `isDeleted`, `isSyncing`, `lastEdited`) lives **only** in the `board` slice (`Board.reducer.js:56-82`). Communicators have no sync state machine: they are reconciled by `lastEdited` inside `syncCommunicators` and pushed as a **side effect** of board push (`updateApiObjectsNoChild` ends with `upsertApiCommunicator`).

"Needs create vs. update" is guessed by id length with **3 different constants** for the same concept:
- `SHORT_ID_MAX_LENGTH = 14` in `Board.constants.js` (used by `isLocalBoard`/`isServerBoard`)
- `SHORT_ID_MAX_LENGTH = 15` hard-coded in `upsertApiCommunicator` (`Communicator.actions.js:63`)
- `LOCAL_COMMUNICATOR_ID` in `api.js`

This asymmetry is the root cause of bug #2203 / PR #2204. The ownership invariant ("fork the default communicator before pushing") had to be copy-pasted to 6 call sites because there was no single sync boundary for the communicator domain.

## Proposed solution

Give communicators their own `syncMeta` in the communicator slice, mirroring the board pattern:

- Explicit `status` (`pending` / `syncing` / `synced`) per communicator id
- "needs create vs. update" becomes explicit state — not a heuristic
- Introduce a value object for server-vs-local identity instead of `id.length` comparison
- The communicator push becomes a first-class operation, not a side effect of board push
- The ownership invariant lives in ONE boundary (`pushCommunicator` thunk, introduced in this branch)

## Why NOT a global syncMeta

Boards, communicators, and settings have **different sync semantics** (conflict resolution strategy, cardinality — settings is a singleton, boards/communicators are collections). A single shared slice couples things that change for different reasons, violating the Common Closure Principle. Each domain owns its sync state.

## Scope / risk

Large refactor, touches the critical login/sync path. Must be backed by regression tests before starting. **Do NOT combine with the hotfix.**

## Blocked by / related

- Fixes #2203
- Depends on: PR #2204 (already merged) + the `pushCommunicator` / mapper fix on this same branch
- This issue should be tackled after the root-cause fix is merged

RodriSanchez1:
[...]

PR #2208, opened 2026-06-01
fix: root-cause fix for communicator sync 500 — mapper + pushCommunicator invariant

### LINKED ISSUE #2203: Sync engine: first board sync after login fails with 500 on PUT /communicator/:id
## Description

After logging in with an account that has no communicator and no boards on the server, doing the first board sync with two or more local boards pending fails with `500 Internal Server Error` on `PUT /communicator/:id`.

The first board push succeeds (POST `/communicator`). The second board push then issues a PUT with a payload that still carries the **default communicator** fields (`email: 'support@cboard.io'`, `author: 'Cboard Team'`) instead of the user's data, which the server rejects.

## Steps to reproduce

1. Create an account that has no communicator nor boards on the server.
2. Before logging in, have at least two local boards pending sync (e.g. created while unlogged or modified default boards).
3. Log in and let the sync engine run (`getApiObjects` → `syncBoards` → `pushLocalChangesToApi`).
4. Observe Network tab: first POST `/communicator` returns 200, second PUT `/communicator/:id` returns 500.

## Root cause

`pushLocalChangesToApi` in `src/components/Board/Board.actions.js` calls `updateApiObjectsNoChild` for each board needing creation. That helper ends with `upsertApiCommunicator(comm)`:

- First call: `comm.id === 'cboard_default'` → goes through `createApiCommunicator` (POST). `api.js` strips the id and overrides `email` / `author` with the user's data, so the POST succeeds.
- The `CREATE_API_COMMUNICATOR_SUCCESS` reducer only copies `id` and `lastEdited` from the response, leaving `email` and `author` as the default values locally.
- Second call: `comm.id` is now the Mongo id from the server → goes through `updateApiCommunicator` (PUT). The PUT body is sent verbatim, still carrying `email: 'support@cboard.io'` and `author: 'Cboard Team'` → server returns 500.

The rest of the app (CommunicatorDialog: `copyBoard`, `setRootBoard`, `deleteMyBoard`, board reorder) avoids this by calling `verifyAndUpsertCommunicator` before any communicator push, which forks the default communicator to one owned by the user (new shortid + user's email / author). The new sync engine path bypasses this step.

## Expected behavior

The sync engine should follow the same invariant as the rest of the app: before pushing the active communicator to the API, ensure it has been forked from the default if it still belongs to `support@cboard.io`.

tomivm:
## Summary

Follows up on #2204 (already merged) with the actual root-cause fix and a structural change to prevent recurrence.

**Root cause:** `CREATE_API_COMMUNICATOR_SUCCESS` was discarding the server's `email` and `author` from the POST response, leaving `support@cboard.io` in local state. The second board's push then sent that stale email via PUT → 500. The server already returned the correct user-owned fields — the mapper just wasn't applying them.

**Structural problem:** The "fork the default communicator before pushing" invariant was replicated across 6 call sites by convention. That's why #2203 happened — the sync engine was the only path that didn't follow the convention.

## Changes (4 atomic commits)

1. **`fix(communicator)` — mapper fix** (`Communicator.reducer.js`)  
   `CREATE_API_COMMUNICATOR_SUCCESS` now merges `email` + `author` from the server response alongside `id` + `lastEdited`. Adds a regression test.

2. **`refactor(communicator)` — `pushCommunicator` thunk** (`Communicator.actions.js`)  
   Single entry point that wraps `verifyAndUpsertCommunicator` + `upsertApiCommunicator`. Future call sites cannot forget the invariant.

3. **`refactor(dialog)` — call site migration** (`CommunicatorDialog.container.js`)  
   `setRootBoard`, `updateCommunicatorBoards`, `deleteMyBoard` → each triple step collapsed into a single `pushCommunicator` call.

4. **`refactor(sync)` — remove redundant email gate** (`Board.actions.js`)  
   The `email !== userEmail` check added by #2204 is already handled inside `verifyAndUpsertCommunicator`. Removed the duplicate.

## Test plan

- [ ] Login to a fresh account (no remote communicator, no boards on server) with two local boards pending sync → both pushes succeed; second `PUT /communicator/:id` returns 200.
- [ ] Login with one board pending (POST-only path) → no regression.
- [ ] Login where local boards only need UPDATE (no CREATE) → communicator not forked.
- [ ] Login where active communicator already belongs to the user (`email === userEmail`) → no fork, behavior unchanged.
- [ ] `CommunicatorDialog` flows (setRoot / reorder / copy / delete board) → unchanged.
- [ ] `yarn test src/components/Communicator src/components/Board` → all green.

## Related

- Fixes #2203
- Builds on #2204
- Tracked architectural follow-up: #2207 (per-domain `syncMeta` for communicators)

### DIFF
--- a/src/components/Communicator/Communicator.reducer.js
+++ b/src/components/Communicator/Communicator.reducer.js
@@ -213,8 +213,7 @@ function communicatorReducer(state = initialState, action) {
 
     case CREATE_API_COMMUNICATOR_SUCCESS:
       // need to check if it was the active communicator as well.
-      // Apply server-authoritative fields (email, author) so subsequent
-      // PUT requests carry the correct owner — not the default communicator's.
+
       return {
         ...state,
         isFetching: false,
--- a/src/components/Communicator/Communicator.reducer.js
+++ b/src/components/Communicator/Communicator.reducer.js
@@ -212,7 +212,9 @@ function communicatorReducer(state = initialState, action) {
       return { ...state };
 
     case CREATE_API_COMMUNICATOR_SUCCESS:
-      // need to check if it was the active communicator as well
+      // need to check if it was the active communicator as well.
+      // Apply server-authoritative fields (email, author) so subsequent
+      // PUT requests carry the correct owner — not the default communicator's.
       return {
         ...state,
         isFetching: false,
@@ -225,6 +227,8 @@ function communicatorReducer(state = initialState, action) {
             ? {
                 ...communicator,
                 id: action.communicator.id,
+                email: action.communicator.email,
+                author: action.communicator.author,
                 lastEdited: action.communicator.lastEdited
               }
             : communicator
--- a/src/components/Board/Board.actions.js
+++ b/src/components/Board/Board.actions.js
@@ -728,14 +728,16 @@ export function pushLocalChangesToApi(remoteBoards = []) {
     });
 
     // If any board needs to be created, ensure the active communicator is owned
-    // by the logged-in user before any push runs.
-    // verifyAndUpsertCommunicator is a no-op when the email already matches.
+    // by the logged-in user before any push runs. Guard on email inequality:
+    // verifyAndUpsertCommunicator re-stamps lastEdited on every dispatch (via
+    // EDIT_COMMUNICATOR), so re-running it on an already-owned communicator
+    // would bump lastEdited and risk a spurious PUT on the next sync.
     if (boardsToSync.some(b => b.needsCreate)) {
       const { communicators, activeCommunicatorId } = getState().communicator;
       const activeCommunicator = communicators.find(
         c => c.id === activeCommunicatorId
       );
-      if (activeCommunicator) {
+      if (activeCommunicator && activeCommunicator.email !== userEmail) {
         dispatch(verifyAndUpsertCommunicator(activeCommunicator));
       }
     }
--- a/src/components/Board/Board.actions.js
+++ b/src/components/Board/Board.actions.js
@@ -727,12 +727,15 @@ export function pushLocalChangesToApi(remoteBoards = []) {
       dispatch
     });
 
+    // If any board needs to be created, ensure the active communicator is owned
+    // by the logged-in user before any push runs.
+    // verifyAndUpsertCommunicator is a no-op when the email already matches.
     if (boardsToSync.some(b => b.needsCreate)) {
       const { communicators, activeCommunicatorId } = getState().communicator;
       const activeCommunicator = communicators.find(
         c => c.id === activeCommunicatorId
       );
-      if (activeCommunicator && activeCommunicator.email !== userEmail) {
+      if (activeCommunicator) {
         dispatch(verifyAndUpsertCommunicator(activeCommunicator));
       }
     }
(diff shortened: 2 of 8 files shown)


src/components/Communicator/Communicator.reducer.js
```diff
@@ -225,6 +227,8 @@ function communicatorReducer(state = initialState, action) {
             ? {
                 ...communicator,
                 id: action.communicator.id,
+                email: action.communicator.email,
+                author: action.communicator.author,
```
RodriSanchez1, line 230: Should we also apply the other server state on the local communicator as boards and rootBoard?

src/components/Communicator/Communicator.reducer.js
```diff
@@ -212,7 +212,9 @@ function communicatorReducer(state = initialState, action) {
       return { ...state };
 
     case CREATE_API_COMMUNICATOR_SUCCESS:
-      // need to check if it was the active communicator as well
+      // need to check if it was the active communicator as well.
+      // Apply server-authoritative fields (email, author) so subsequent
+      // PUT requests carry the correct owner — not the default communicator's.
```
RodriSanchez1: this comment is not needed

RodriSanchez1 (changes requested):
Good work!!  Check this out

 ⚠️ Finish the migration: apply pushCommunicator everywhere the old pattern survives (commits 2–3)

  The abstraction is the right call, but it's only half-applied. The stated goal — "single call site for every communicator push — the invariant
  cannot be forgotten" — isn't met yet, because two call sites still use the old verify + upsertApi sequence by hand:

  - CommunicatorToolbar.editCommunicatorTitle (CommunicatorToolbar.container.js:50)
  - Import.addBoardsToCommunicator (Import.container.js:154)

  Both are byte-for-byte the pattern pushCommunicator replaces:

```js
  const upserted = verifyAndUpsertCommunicator(data);
  if ('name' in userData && 'email' in userData) {
    try { await upsertApiCommunicator(upserted); } catch (err) {...}
  }
```
  Please migrate these two to pushCommunicator as well, so the invariant is actually centralized and we don't end up with two competing idioms.


  ❌ Drop: the gate removal (commit 4)

  The comment "verifyAndUpsertCommunicator is a no-op when the email already matches" is inaccurate. In the matched case it still dispatches
  editCommunicator, which re-stamps lastEdited to now on every call (and mutates the communicators array in place). Since lastEdited drives sync
  decisions, removing the `email !== userEmail` guard means every sync  wit-needsCreate re-stamps an already-owned communicator and can trigger
  a spurious PUT on the next sync — the opposite of what we want. The cheap equality guard is the safer choice; I'd keep it.

src/components/Communicator/Communicator.reducer.js
tomivm: What you mean?

RodriSanchez1: The POST response returns the entire communicator object, which contains more properties than just `author`, `email`, `lastEdited`, and `id`. For example, it also includes fields such as `rootBoard` and `boards[]`.

 This change already applies `email: action.communicator.email`, but maybe  the same approach should be applied to the rest of the communicator fields to keep the local state fully synchronized with the server response.

tomivm: We can analyse it in a separate issue

PR #2209, opened 2026-06-01
Clean PIN in navigationSettings on logout

magush27:
Clean PIN in `navigationSettings` on logout to prevent the error of showing the PIN of another account after it's logout.

### DIFF
--- a/src/components/App/App.reducer.js
+++ b/src/components/App/App.reducer.js
@@ -167,7 +167,12 @@ function appReducer(state = initialState, action) {
     case LOGOUT:
       return {
         ...state,
-        userData: {}
+        userData: {},
+        navigationSettings: {
+          ...state.navigationSettings,
+          pinLockEnabled: false,
+          pinCode: ''
+        }
       };
     case UPDATE_USER_DATA:
       return {
--- a/src/components/App/__tests__/App.reducer.test.js
+++ b/src/components/App/__tests__/App.reducer.test.js
@@ -109,6 +109,26 @@ describe('reducer', () => {
     };
     expect(appReducer(initialState, logout)).toEqual(initialState);
   });
+  it('should clear PIN on logout', () => {
+    const stateWithPin = {
+      ...initialState,
+      userData: uData,
+      navigationSettings: {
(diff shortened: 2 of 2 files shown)


RodriSanchez1:
The existing logout test doesn't actually exercise the fix (minor but worth addressing).
App.reducer.test.js:106 runs LOGOUT against initialState, where pinLockEnabled is already false and pinCode is already ''. The assertion toEqual(initialState) passes both before and after this PR — so there's no regression protection for the new behavior. Consider adding a case that starts from a state with a PIN set:

```js
it('should clear PIN on logout', () => {
  const stateWithPin = {
    ...initialState,
    userData: uData,
    navigationSettings: {
      ...initialState.navigationSettings,
      pinLockEnabled: true,
      pinCode: '1234'
    }
  };
  expect(appReducer(stateWithPin, { type: LOGOUT })).toEqual({
    ...stateWithPin,
    userData: {},
    navigationSettings: {
      ...stateWithPin.navigationSettings,
      pinLockEnabled: false,
      pinCode: ''
    }
  });
});
```

magush27:
> The existing logout test doesn't actually exercise the fix (minor but worth addressing). App.reducer.test.js:106 runs LOGOUT against initialState, where pinLockEnabled is already false and pinCode is already ''. The assertion toEqual(initialState) passes both before and after this PR — so there's no regression protection for the new behavior. Consider adding a case that starts from a state with a PIN set:
> 
> ```js
> it('should clear PIN on logout', () => {
>   const stateWithPin = {
>     ...initialState,
>     userData: uData,
>     navigationSettings: {
>       ...initialState.navigationSettings,
>       pinLockEnabled: true,
>       pinCode: '1234'
>     }
>   };
>   expect(appReducer(stateWithPin, { type: LOGOUT })).toEqual({
>     ...stateWithPin,
>     userData: {},
>     navigationSettings: {
>       ...stateWithPin.navigationSettings,
>       pinLockEnabled: false,
>       pinCode: ''
>     }
>   });
> });
> ```

Awesome! Already solved it in the last commit, thanks

RodriSanchez1 (approved):

Issue #2195, opened 2026-05-26
[Epic] Sync engine: cascade failure resilience and silent inconsistency gaps


RodriSanchez1:
## Summary

The board sync engine (see `docs/sync-engine.md`) is mostly resilient at the **per-board** level — each push/delete is wrapped in its own try/catch, so a single failure doesn't stop the rest of the batch and failed boards remain `PENDING` for retry. However, an in-depth review of the **cascading side effects** triggered by `updateApiObjectsNoChild` → `CREATE_API_BOARD_SUCCESS` → `updateApiMarkedBoards` surfaced a class of failures where:

- Local state is mutated optimistically (IDs swapped, `markToUpdate`/`shouldCreateBoard` flags set, communicator updated locally).
- A subsequent API call in the cascade fails (`upsertApiCommunicator`, or `updateApiBoard` inside `updateApiMarkedBoards`).
- The board's `syncMeta` is already `SYNCED`, so the standard PENDING-driven retry path **does not pick it up**.
- The only recovery path is *another* successful create cascade that re-runs `updateApiMarkedBoards`.

The net effect is **silent divergence between local and server state**, invisible to the user because `SYNC_BOARDS_SUCCESS` is dispatched regardless of partial cascade failures.

## Scope

This epic groups all the gaps found in that analysis. They fall into three buckets:

1. **Observability** — partial failures must be surfaceable to the UI.
2. **Tracking** — every piece of state that needs to be pushed must live in `syncMeta`, not in ad-hoc board flags.
3. **Cascade consistency** — error handling and ordering inside the cascade need to be uniform and recoverable.

## Sub-issues

- [ ] #2196 — Surface partial failures via `syncError` instead of silent `SYNC_BOARDS_SUCCESS`
- [ ] #2197 — `upsertApiCommunicator` failures leave server-side communicator desynced silently
- [x] #2198 — `markToUpdate` must be reflected in `syncMeta` so Pass 1 retries stale tile references
- [ ] #2199 — Asymmetric error handling in `updateApiMarkedBoards` orphans batch on `markToUpdate` failure
- [ ] #2200 — `replaceHistoryWithActiveBoardId` skipped if cascade throws after `CREATE_API_BOARD_SUCCESS`
- [ ] #2201 — `upsertApiCommunicator` called per cascade level — coalesce to a single push per sync
- [ ] #2202 — Recursive `updateApiObjectsNoChild` ↔ `updateApiMarkedBoards` has no depth bound or cycle detection

[...]

PR #2219, opened 2026-06-09
fix(sync): keep board PENDING while it references a not-yet-created child (#2218)

### LINKED ISSUE #2218: Sync: server can persist a board referencing a local (short) board id
Part of #2195 (sync engine resilience epic). **Priority: High** — produces persistent server-side data corruption with no automatic recovery.

## Summary

When local board changes are synced to the API, a board can be saved on the server with a tile whose `loadBoard` points to a **short, locally-generated id** (`shortid`) instead of a server-assigned board id. That short id does not correspond to any board on the server, so the reference is **dangling**: opening that folder tile from a board pulled on another device (or after a fresh pull) resolves to nothing.

The server state ends up **out of sync** with local state: locally the parent correctly points at the server child, but the copy persisted on the server still points at the old short id.

## Background

The client generates board ids locally with `shortid.generate()`. When a board is created on the server (`POST /board`), the server assigns a new (MongoDB ObjectId-style) id and returns it. Any reference to the old short id must then be rewritten everywhere it appears:

- `tile.loadBoard` on parent boards
- `communicator.boards[]`, `communicator.rootBoard`, `communicator.activeBoardId`

The interactive edit path (`updateApiObjects`, `Board.actions.js:1152`) handles this atomically: it creates the child, reads back the new id, rewrites the parent's `tile.loadBoard` (lines 1163-1168), and pushes the parent in the same operation.

The **sync path** (`pushLocalChangesToApi`, `Board.actions.js:747`) does not. It pushes boards one at a time via `updateApiObjectsNoChild`, and the parent's `loadBoard` is only reconciled by a separate, deferred best-effort step (`CREATE_API_BOARD_SUCCESS` rewriting local state + `markToUpdate` / `shouldCreateBoard` re-push via `updateApiMarkedBoards`). That second step is not atomic with the first push, so it can be lost.

## Why it happens

A folder created while **logged out / offline** never goes through the atomic edit path (`handleApiUpdates` early-returns when there is no logged-in user, `Board.container.js:1087`). It produces purely local state:

- `CREATE_BOARD` → child board `C` with a short id, marked `PENDING`, **appended to the end** of `state.boards` (`Board.reducer.js:298`).
- `CREATE_TILE` on parent `P` → tile with `loadBoard = C.shortId`, `P` marked `PENDING` (`Board.reducer.js:328`).

On the next sync, `pushLocalChangesToApi` pushes boards in `state.boards` array order (`Board.actions.js:688`). Because the new child was appended last, **the parent is pushed before the child**:

1. Parent `P` is pushed first, while `C` still has its short id, so the request body carries `loadBoard = C.shortId`. The server now stores a dangling reference.
2. Child `C` is created next; `CREATE_API_BOARD_SUCCESS` (`Board.reducer.js:379`) rewrites `P` in **local state** and re-marks `P` `PENDING`.
3. A deferred pass (`updateApiMarkedBoards`) is expected to re-push the corrected `P`.

Step 1 always happens. Step 3 is a separate, later request — it is what is supposed to fix the server, and it can be dropped.

`REPLACE_BOARD` (`Board.reducer.js:216`), which the sync path dispatches, does **not** rewrite parent `loadBoard` references.

## Impact

- **Transient (common):** on essentially every sync that materializes an offline-created folder, the server briefly holds the parent with a short `loadBoard` between step 1 and step 3.
- **Persistent (data corruption):** the corrected re-push (step 3) can fail to land while the parent push (step 1) already succeeded, leaving the server permanently pointing at a non-existent short id. Known triggers:
  - The child create fails (validation/network) while the parent push succeeded. The parent is now `SYNCED` on the server with the short id; it is only re-marked `PENDING` if/when the child is eventually created, so if the child keeps failing the dangling reference is permanent.
  - The sync run is interrupted (tab closed, app killed, connection dropped) between the parent push and the child create / corrected re-push.
  - The deferred re-push is silently dropped — errors in the push loop are swallowed (`Board.actions.js:804`), and `updateApiMarkedBoards` aborts the whole pass with `return` instead of `continue` when a board id changed mid-loop (`Board.actions.js:1098`).

Result on affected accounts: a folder tile that opens correctly on the device that created it, but opens to an empty/missing board on any other device or after the local cache is cleared, because the server copy references an id that was never saved.

### Manual (real-world)

1. Log in as a user (so the parent is a real server board with a long id).
2. Open DevTools → Network → set to **Offline**.
3. Inside one of your boards, **add a folder tile**. This creates child board `C` with a short id; the parent `P` now has a tile with `loadBoard = C.shortId`; both are marked `PENDING`. Edit a tile inside the new folder so `C` is dirty.
4. Go back **Online**. Sync starts and pushes parent `P` first, carrying `C.shortId`.
5. Force the child to not complete: use DevTools **"Block request URL"** on the child `POST /board` (or switch back to Offline) right after the parent's request completes.
6. Inspect `P` on the server (fetch it directly, or open it on a second device): its folder tile's `loadBoard` is the short id, and opening that folder elsewhere resolves to nothing.

> If the child create is allowed to succeed, the Network tab still shows the bad first push (parent with the short id) followed by a corrected push — that is the self-heal. The persistent corruption requires the corrected push to be lost (step 5).

### Deterministic (automated)

Using the real reducer + thunk with the `../../api` module mocked, capture the request payloads and force the child create to reject:

```js
import API from '../../api';
jest.mock('../../api');

it('server receives a tile.loadBoard pointing at a local short id', async () => {
  const parent = {
    id: '1234567890ABCDEF',                 // long id = existing server board
    email: 'me@x.com',
    tiles: [{ id: 't1', loadBoard: 'childShort' }],
  };
  const child = { id: 'childShort', email: 'me@x.com', tiles: [] }; // needs create

  const store = makeRealStore({
    boards: [parent, child],
    syncMeta: {
      '1234567890ABCDEF': { status: 'PENDING' },
      childShort: { status: 'PENDING' },
    },
  });

  const puts = [];
  API.updateBoard.mockImplementation(b => { puts.push(b); return Promise.resolve(b); });
  API.createBoard.mockRejectedValue(new Error('boom')); // child create fails

  await store.dispatch(actions.pushLocalChangesToApi([]));

  // Parent was sent to the server carrying the LOCAL short id:
  expect(puts[0].tiles[0].loadBoard).toBe('childShort');
  // and it is never corrected, because the child never got a server id:
  expect(puts.some(p => p.tiles[0].loadBoard !== 'childShort')).toBe(false);
});
```

Removing the `mockRejectedValue` line lets the test observe the self-heal (a second, corrected push), which is why the bug is intermittent rather than constant.

## Relevant code

- `pushLocalChangesToApi` — sync push, per-board, parent-before-child order: `Board.actions.js:747`
- Push loop, swallowed errors: `Board.actions.js:786`
- `updateApiObjectsNoChild` (no parent rewrite): `Board.actions.js:1045`
- `updateApiMarkedBoards` `return`-instead-of-`continue`: `Board.actions.js:1098`
- `updateApiObjects` (atomic edit path, for contrast): `Board.actions.js:1152`
- `CREATE_API_BOARD_SUCCESS` local rewrite: `Board.reducer.js:379`
- `REPLACE_BOARD` (does not rewrite parent refs): `Board.reducer.js:216`
- Offline creation marks boards `PENDING`: `Board.reducer.js:298`, `:328`
- Edit path early-returns when logged out: `Board.container.js:1087`

RodriSanchez1:
## Summary

Fixes #2218. When local board changes are synced to the API, a parent board could be persisted on the server with a tile whose `loadBoard` points to a **local (short) child id** that does not exist on the server — a permanently dangling reference. The folder opens correctly on the device that created it but resolves to nothing on any other device or after a fresh pull.

The root cause is that the sync path graduates a board to `SYNCED` as soon as its own push succeeds, even when the child it links to has not been created yet. If the child create later fails (or the corrected re-push is dropped), the server is left pointing at a short id forever.

## Fix

Introduce `hasUnsyncedChildReference(board, boards)` and gate **both** sync graduation points on it — do not mark a board `SYNCED` while it still references a local, not-yet-created child board; keep it `PENDING` so the next sync run retries instead of locking in a broken reference.

- **`UPDATE_API_BOARD_SUCCESS`** — covers existing server (long-id) parents (the scenario in the issue).
- **`CREATE_API_BOARD_SUCCESS`** — covers parents that are *created* rather than updated. This includes **transformed default boards**: modifying a default board (e.g. adding a folder) transforms it to a user board and creates it on the server, so it lands entirely on the create path and never hit the update-path guard. This is the most common trigger.

Default boards are excluded from the check, since they are intentionally never pushed (kept local with short ids) — a link to one is not a pending reference and must not trap the parent in `PENDING`.

## Tests

Added reducer regression tests:
- Keep parent `PENDING` on `UPDATE_API_BOARD_SUCCESS` while it references a local child.
- Keep the just-created board `PENDING` on `CREATE_API_BOARD_SUCCESS` while it references a local grandchild (create path).
- Graduate to `SYNCED` when the reference points at a server child.
- Graduate to `SYNCED` when the reference points at a default board (never trapped in `PENDING`).

## Notes / scope

This is the reducer-level mitigation: it prevents `SYNCED`-with-corruption. In the permanent-failure case the parent now stays `PENDING` and keeps retrying, but each retry still re-pushes the short `loadBoard` until the child finally gets a server id. Fully eliminating the bad push (ordering children before parents in the push loop) is the actions-layer part of the resilience epic (#2195) and is out of scope here.


magush27:
It looks good!
I have a suggestion about a problem with `createBoard` that can be undefined, for example, if the board was deleted from state between the API call and the success response. In this case `hasUnsyncedChildReference` handles it safely returning false, but then `setSyncMeta` puts it as SYNCED and adds a syncMeta entry for a board that no longer exists in boards. It can leave a ghost state. This also happends with UPDATE_API_BOARD_SUCCESS and pushedBoard. I suggest to skip the setSyncMeta call if the board is no longer in state, something similar to this:

```
  const createdBoard = finalBoards.find(b => b.id === action.board.id);                                                                                                                            
  const createdStillPending = hasUnsyncedChildReference(createdBoard, finalBoards);                                                                                                                
  return {                                                                                                                                                                                         
    ...state,                
    isFetching: false,
    boards: finalBoards,
    syncMeta: createdBoard
      ? setSyncMeta(newSyncMeta, action.board.id, {
          status: createdStillPending ? SYNC_STATUS.PENDING : SYNC_STATUS.SYNCED
        })
      : newSyncMeta   // board was deleted before the response arrived, don't leave ghost state
  };
```

And for `UPDATE_API_BOARD_SUCCESS` something like this:
```
  const pushedBoard = state.boards.find(b => b.id === action.boardData.id);
  const stillPending = hasUnsyncedChildReference(pushedBoard, state.boards);
  return {
    ...state,
    isFetching: false,
    boards: state.boards.map(...),
    syncMeta: pushedBoard
      ? setSyncMeta(state.syncMeta, action.boardData.id, {
          status: stillPending ? SYNC_STATUS.PENDING : SYNC_STATUS.SYNCED
        })
      : state.syncMeta
  };
```

This way we can keep state clean in both cases.

Other thing I've noticed is about tests, that they have a semantic flaw: `mockBoard` has `id: '123'` but the test puts `{ ...mockBoard, id: '456456456456456456456' }` in state and uses `boardId: '456456456456456456456'` (21 chars, ≥ SHORT_ID_MAX_LENGTH). That makes `isLocalBoard()` classify the board being "created" as a server board, which means Pass 1's boardsToMarkForCreation logic never fires for it. `CREATE_API_BOARD_SUCCESS` is supposed to fire for local boards getting their first server id, so `boardId` should be a short id like '123' and not a server-length one.

Also, I noticed there's no unit tests for `hasUnsyncedChildReference`. It's only executed through the reducer, which is indirect. Edge cases like boards with no tiles, boards being empty or a tile with `loadBoard: '' ` aren't explicitly documented here or tested even though the implementation handles them. 

Other thing that hasn't got tests is a full resolution cycle. The new tests cover parents staying PENDING, but there's no test for the resolution path where child gets its server id (fires `CREATE_API_BOARD_SUCCESS` for the child, which in Pass 2 updates the parent's tile `loadBoard` to the server id) → parent's next `UPDATE_API_BOARD_SUCCESS` fires → `hasUnsyncedChildReference` now returns false → parent graduates to SYNCED.

PR #2240, opened 2026-06-26
fix: single-flight guard for the sync run

### LINKED ISSUE #2238: Sync can create a board twice when triggers fire concurrently
## Problem

Two sync triggers firing close together can run `getApiObjects()` concurrently. When a board is `PENDING`/`needsCreate`, both runs read the same pre-PUSH snapshot and both POST the board → **the board is created twice on the server**.

Root cause: there is no in-flight guard on the sync pipeline. The only gate — the time throttle in `App.container.js` (`isSyncRecentlyExecuted`) — is deliberately bypassed when `hasPendingSyncBoards` is true, which is exactly the case where a double-create is harmful.

There are **two** entry points into `getApiObjects`:

- `App.container.js → handleDataRefresh` (the 5 automatic triggers).
- `SyncButton.js → onSyncClick` (manual sync).

A component-level guard only protects the first door, so a manual sync overlapping an automatic trigger can still double-fire. The guard must live at the shared chokepoint: `getApiObjects`.

## Fix

Add a single-flight guard at `getApiObjects` driven by the existing `isSyncing` state, raised **before the first `await`** (race-free, since redux `dispatch` is synchronous) and cleared in a `.finally` on every exit path (leak-free). This separates two concepts currently conflated in `SYNC_BOARDS_*`:

- **Sync run** (boards + communicators) → gated by `isSyncing` via new `SYNC_STARTED` / `SYNC_FINISHED`.
- **Board sync** (PULL/PUSH phase) → keeps `syncError` + telemetry via `SYNC_BOARDS_*`.

RodriSanchez1:
Closes #2238

## Problem

Two sync triggers firing close together can run `getApiObjects()` concurrently. When a board is `PENDING`/`needsCreate`, both runs read the same pre-PUSH snapshot and both POST the board → **the board is created twice on the server**.

The only existing gate — the time throttle in `App.container.js` — is deliberately bypassed when `hasPendingSyncBoards` is true, which is exactly the harmful case. And there are two entry points into `getApiObjects` (`handleDataRefresh` and `SyncButton`), so a component-level guard can't cover both.

## Fix

Move the guard to the shared chokepoint, `getApiObjects`, driven by the existing `isSyncing` state:

- No-ops when `getState().board.isSyncing` is already `true`.
- Dispatches `syncStarted()` **before the first `await`** — race-free, since redux `dispatch` is synchronous.
- Dispatches `syncFinished()` in `.finally` on **every** exit path — leak-free, so a failed boards/communicators fetch can't leave the flag stuck.

This separates two concepts previously conflated in `SYNC_BOARDS_*`:

| Concept | State | Driven by |
| --- | --- | --- |
| Sync run (boards + communicators) | `isSyncing` | new `SYNC_STARTED` / `SYNC_FINISHED` |
| Board sync (PULL/PUSH phase) | `syncError` + telemetry | `SYNC_BOARDS_*` (slimmed to `syncError` only) |

## Impact

- `isSyncing` is read only by `SyncButton.getSyncStatus` — the spinner now spans both the boards and communicators phases (arguably more correct, no regression).
- `syncError` has no consumer; kept board-scoped.
- Telemetry is unchanged (direct calls inside `syncBoards`).

## Tests

- `getApiObjects` thunk: no-ops when `isSyncing` is already true; `SYNC_STARTED`→`SYNC_FINISHED` on success; `SYNC_FINISHED` always fires on the failure path and the non-array-response path (stuck-flag guard).
- Reducer: new `SYNC_STARTED`/`SYNC_FINISHED` cases; `SYNC_BOARDS_*` no longer touch `isSyncing`.
- `App.container`: throttle-bypass and throttled coverage replacing the old in-flight component tests.

All 160 affected tests pass.

### DIFF
--- a/src/components/App/App.container.js
+++ b/src/components/App/App.container.js
@@ -17,7 +17,7 @@ import {
   updateUnloggedUserLocation,
   updateConnectivity
 } from '../App/App.actions';
-import { getApiObjects } from '../Board/Board.actions';
+import { getApiObjects, clearSync } from '../Board/Board.actions';
 import {
   isCordova,
   isElectron,
@@ -56,7 +56,11 @@ export class AppContainer extends Component {
     /**
      * Get API objects (boards and communicators)
      */
-    getApiObjects: PropTypes.func.isRequired
+    getApiObjects: PropTypes.func.isRequired,
+    /**
+     * Clear a stale isSyncing flag rehydrated from a previous session
+     */
+    clearSync: PropTypes.func.isRequired
   };
 
   lastSyncTime = null;
@@ -138,6 +142,7 @@ export class AppContainer extends Component {
       this.handleWebVisibilityChange
     );
 
+    this.props.clearSync();
     this.handleDataRefresh('App started');
   }
 
@@ -264,7 +269,8 @@ const mapDispatchToProps = {
   updateLoggedUserLocation,
   updateUnloggedUserLocation,
   updateConnectivity,
-  getApiObjects
+  getApiObjects,
+  clearSync
 };
 
 export default connect(
--- a/src/components/App/App.container.js
+++ b/src/components/App/App.container.js
@@ -152,7 +152,7 @@ export class AppContainer extends Component {
   }
 
   isSyncRecentlyExecuted = () => {
-    const THROTTLE_MS = 1000 * 60 * 2;
+    const THROTTLE_MS = 1000 * 30;
     return this.lastSyncTime && Date.now() - this.lastSyncTime < THROTTLE_MS;
   };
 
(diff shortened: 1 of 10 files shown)


src/components/App/App.container.js
```diff
@@ -152,7 +152,7 @@ export class AppContainer extends Component {
   }
 
   isSyncRecentlyExecuted = () => {
-    const THROTTLE_MS = 1000 * 60 * 2;
+    const THROTTLE_MS = 1000 * 30;
```
RodriSanchez1, line 160: The throttle time was reduced to 30 secs instead of two minutes

PR #2244, opened 2026-06-28
fix: cache remote symbol images on-device for offline use

### LINKED ISSUE #2243: Global Symbols images don't show when offline
GlobalSymbols images are loaded straight from globalsymbols.com (e.g. `https://globalsymbols.com/uploads/production/image/imagefile/...`) every time they're shown.

When the device has no internet, the image fails and the user just sees the broken/blank image-not-found placeholder. That's useless for someone relying on the board to communicate.

We should cache these images on the device so they keep working offline, the same way we already handle other symbol sets.

eturino:
Closes #2243

GlobalSymbols images were loaded straight from globalsymbols.com on every render, so they showed a blank image-not-found when the device was offline.

## What this does
- Caches remote (`http`/`https`) symbol images in IndexedDB, keyed by URL.
- In the `Symbol` component: if the image is already cached, render the local copy. If not and we're online, show the network image now and save it in the background for next time. If offline and not cached, fall back to the current behavior.

## Why IndexedDB and not the service worker cache
The service worker doesn't run in the native (Cordova) app, so it wouldn't help there. IndexedDB works on both web and native.

Keying by URL means this also covers any other remote symbol image, not just GlobalSymbols.

## Caching is opt-in per call site

`Symbol` only *writes* to the cache when given `cacheRemoteImage`. Reads are unconditional, so an image cached earlier still renders offline everywhere.

It is set only where an image is already committed to a board - `Board` and `SymbolOutput` - and deliberately left off in `SymbolSearch` and in the `TileEditor` preview.

The editor preview is worth calling out: `handleSymbolSearchChange` updates `tileInView.image` the moment a suggestion is *picked*, so caching there would store every symbol tried and abandoned. Saving instead routes through `editTiles`, after which `Board` renders the tile and caches it - so a kept tile is still cached, without the discarded ones.

Without this, every symbol search would permanently store up to 20 GlobalSymbols images (`limit=20` per query, plus ARASAAC results), whether or not the user picked any of them. Searching is unbounded, so the cache would be too, and it shares the origin's storage quota with the `cboard` database that holds the user's actual boards.

Those entries also have no offline value: GlobalSymbols suggestions only appear after a live API call, so offline there are no suggestions to render them for.

Default-off was chosen over default-on so that a new list-style view can't silently reintroduce unbounded caching. The trade-off is that a new tile call site that forgets the prop loses offline support - a quieter failure, but it degrades to today's behavior rather than to a storage problem.

## Storage budget

`putCachedImage` refuses to write past `min(250MB, 10% of quota)`, measured with `navigator.storage.estimate()`. Over budget it stops caching and the symbol loads from the network - the behavior before this PR - rather than failing.

Quota is granted out of free disk, so the share scales down on a device with little space left. Measured in the Cordova WebView on a Pixel 10a: quota 10.7 GB, `isSecureContext: true`, `estimate()` available. Where it isn't available the budget check is skipped rather than blocking the cache.

The absolute cap matters because the boards are the only irreplaceable data in this origin and they are small - 164 KB total on that device - so an unbounded image cache would quickly become the entire storage footprint. Note the ARASAAC offline pack is opt-in and most users never download it, so this cache cannot assume it is the smaller consumer.

## Testing

Verified on a Pixel 10a (Android 17) running the Cordova build, and in a browser.

**On device, offline.** Loaded boards with remote symbols while online, enabled airplane mode, force-stopped the app, and reopened it. All symbols still rendered. Inspected the WebView via the DevTools protocol to confirm it wasn't a stale render:

- `navigator.onLine: false`, origin `file://` (so no service worker involved)
- 43 images cached, 781 KB, across `globalsymbols.com` (26), `static.arasaac.org` (12) and `cdncboard.azureedge.net` (5)
- All 24 rendered symbols served from `blob:` URLs — 0 network, 0 broken
- Stored content types were `image/png` and `image/svg+xml` only

Because the cache keys on any `https://` image, this also covers remote ARASAAC symbols and user-uploaded CDN images, not just GlobalSymbols.

**On device, opt-in boundary.** From a cleared app (0 cached entries): two GlobalSymbols searches left the cache at **0**, with 20 suggestion images rendering from the network. Then choosing a result and saving the tile cached **exactly that one** (`image/png`, 41,439 bytes), and the tile re-rendered from a `blob:` URL. Before this prop, those searches alone would have stored up to 40 images.

**In a browser.** Confirmed the first load fetches and stores the image, and a reload serves it from IndexedDB as a `blob:`.

**Captive portals.** Testing happened on public wifi, which exposed a failure mode: a portal answers `200 OK` with its own HTML, so the response would have been cached as the image and, since a cache hit never re-fetches, that URL would render broken permanently. The upload path is now guarded on `content-type` starting with `image/`, with a unit test covering it.

**CSP.** `connect-src` allows `https://*`, so `fetch()` of remote symbols is permitted. Note it does not allow plain `http://` — such an image fails the fetch, gets caught, and falls back to the network `src`, i.e. today's behavior with no caching.

### DIFF
--- a/src/idb/imageCache/imageCache.ts
+++ b/src/idb/imageCache/imageCache.ts
@@ -1,0 +1,32 @@
+import { DBSchema, openDB } from 'idb';
+
+export interface CachedImage {
+  url: string;
+  type: string;
+  data: ArrayBuffer;
+}
+
+interface ImageCacheDB extends DBSchema {
+  images: {
+    key: string;
+    value: CachedImage;
+  };
+}
+
+const dbPromise = openDB<ImageCacheDB>('cboard-image-cache', 1, {
+  upgrade(db) {
+    db.createObjectStore('images', { keyPath: 'url' });
+  },
+});
+
+export async function getCachedImage(
+  url: string,
+): Promise<CachedImage | undefined> {
+  const db = await dbPromise;
+  return db.get('images', url);
+}
+
+export async function putCachedImage(image: CachedImage): Promise<void> {
+  const db = await dbPromise;
+  await db.put('images', image);
+}
--- a/src/idb/imageCache/imageCache.ts
+++ b/src/idb/imageCache/imageCache.ts
@@ -4,12 +4,17 @@ export interface CachedImage {
   url: string;
   type: string;
   data: ArrayBuffer;
+  lastUsed: number;
 }
+
+// callers cache an image, they don't decide when it was last used
+export type ImageToCache = Omit<CachedImage, 'lastUsed'>;
 
 interface ImageCacheDB extends DBSchema {
   images: {
     key: string;
     value: CachedImage;
+    indexes: { byLastUsed: number };
   };
   meta: {
     key: string;
@@ -21,17 +26,33 @@ const TOTAL_BYTES_KEY = 'totalBytes';
 
 const dbPromise = openDB<ImageCacheDB>('cboard-image-cache', 1, {
   upgrade(db: IDBPDatabase<ImageCacheDB>): void {
-    db.createObjectStore('images', { keyPath: 'url' });
+    const images = db.createObjectStore('images', { keyPath: 'url' });
+    images.createIndex('byLastUsed', 'lastUsed');
     db.createObjectStore('meta');
   }
 });
+
+// a write per render would cost more than the eviction order is worth, so age the
+// timestamp coarsely: a symbol used at all today is as recent as any other
+const TOUCH_AFTER_MS = 24 * 60 * 60 * 1000;
 
 export async function getCachedImage(
   url: string
 ): Promise<CachedImage | undefined> {
   try {
     const db = await dbPromise;
-    return await db.get('images', url);
+    const cached = await db.get('images', url);
+
+    if (cached && Date.now() - cached.lastUsed > TOUCH_AFTER_MS) {
+      try {
+        await db.put('images', { ...cached, lastUsed: Date.now() });
+      } catch (error) {
+        // a failed touch costs eviction order, not the hit we already have
+        console.error('Failed to touch cached image:', error);
+      }
+    }
+
+    return cached;
   } catch (error) {
     console.error('Failed to read cached image:', error);
     return undefined;
@@ -40,16 +61,16 @@ export async function getCachedImage(
 
 const MAX_CACHE_BYTES = 250 * 1024 * 1024;
 const MAX_QUOTA_SHARE = 0.1;
+// free some headroom when evicting, so a full cache doesn't evict on every write
+const EVICT_TO_SHARE = 0.95;
 
-let warnedAboutBudget = false;
+let warned = false;
 
 // Boards are the only irreplaceable thing in this origin and they are tiny (~100s
 // of KB), so an unbounded image cache would be the whole storage footprint. The
 // absolute cap is what enforces that: storage.estimate() only lowers it, and is
 // missing on iOS 16 and old Android WebViews. The browser grants quota out of free
 // disk, so taking a share of it self-limits on a device with little space left.
-// Over budget we simply stop caching: symbols fall back to loading from the
-// network, which is what they did before any of this.
 async function budgetBytes(): Promise<number> {
   const { quota } = (await navigator.storage?.estimate?.()) ?? {};
   return quota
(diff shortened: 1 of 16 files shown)


eturino:
@martinbedouret @shayc I'm not sure who is the best for reviewing this one :) Happy to answer anything and amend whatever is needed. The bug is hitting my son's tablet all the time so I am quite keen on fixing it :D

src/idb/imageCache/imageCache.ts
```diff
@@ -0,0 +1,77 @@
...
+const MAX_QUOTA_SHARE = 0.1;
+
+let warnedAboutBudget = false;
+
+// Boards are the only irreplaceable thing in this origin and they are tiny (~100s
+// of KB), so an unbounded image cache would be the whole storage footprint. The
+// browser grants quota out of free disk, so taking a share of it self-limits on a
+// device with little space left. Over budget we simply stop caching: symbols fall
+// back to loading from the network, which is what they did before any of this.
+async function isWithinBudget(bytes: number): Promise<boolean> {
```
shayc: [P1] The cache becomes unbounded on a supported platform.

[isWithinBudget()](https://github.com/cboard-org/cboard/blob/3224d715d3ec4657fb6db47de578f3dfc993f7a6/src/idb/imageCache/imageCache.ts#L44) returns true whenever navigator.storage.estimate() is unavailable. Cboard supports iOS 16, while WebKit added StorageManager.estimate() in iOS 17. With no eviction, iOS 16 caches until IndexedDB quota failure. Since Cboard’s boards also live in origin-scoped IndexedDB, this can eventually prevent board changes from persisting. [WebKit Storage API availability](https://webkit.org/blog/14403/updates-to-storage-policy/), [Cboard Cordova config](https://github.com/cboard-org/ccboard/blob/master/config.xml)

The size limit must work independently of navigator.storage.estimate(): maintain cache byte totals inside IndexedDB and use estimate() only to lower the limit when available. This fallback also matters for old Android WebViews; Cboard’s Android minimum does not enforce a minimum WebView version.

shayc:
@eturino thank you for the contribution! I've left one comment. I would also wait on @magush27 or @RodriSanchez1 to review it before merging.

I see you got some merge conflicts related to a recent Prettier formatting on `master`.

eturino:
> @eturino thank you for the contribution! I've left one comment. 

I can't see it :( ... is it github being github, or maybe you didn't publish the review?

> I would also wait on @magush27 or @RodriSanchez1 to review it before merging.
> 
> I see you got some merge conflicts related to a recent Prettier formatting on `master`.

Yep, on it!

shayc:
> is it github being github, or maybe you didn't publish the review?

It's me being me 🫤, I forgot to submit... done now.

src/idb/imageCache/imageCache.ts
eturino: good one.

estimate() is now an enhancement, we fall back to the static max. Also keeping track in `meta`.

eturino: Replaced the stop-caching-when-full behaviour with LRU eviction

images carry a lastUsed timestamp (refreshed on read at most once a day), and an over-budget write evicts oldest-first down to 95% of the budget. 

an image bigger than the whole budget is skipped instead of clearing the cache to fit something that never will.

eturino:
@shayc ready for your eyes again :)

eturino:
@shayc @magush27 @RodriSanchez1  I am sorry to bother again, I am quite keen on having it fixed for my son soon, where he's using the tablet these days there is almost always no connectivity

RodriSanchez1:
@eturino Hello! That's an awesome work!! I'am reviewing it.  I'm thinking on maybe extend a bit the db to media, to store sounds, images, and not images only.  I'm on it

eturino:
> @eturino Hello! That's an awesome work!! I'am reviewing it. I'm thinking on maybe extend a bit the db to media, to store sounds, images, and not images only. I'm on it

@RodriSanchez1 that's a good idea! I would make that maybe opt-in though, in case it bloats the available space. In the end, images are more important for this than the sound, imho.

RodriSanchez1:
> that's a good idea! I would make that maybe opt-in though, in case it bloats the available space. In the end, images are more important for this than the sound, imho.

Fair concern, but I think I explained the idea badly — the sounds I have in mind aren't remote ones we'd be caching. They're the images and sounds a user records or uploads **while offline**, which need to live somewhere until the connection comes back and they can be synced.

Today those are kept as base64 in the redux store, which is really the wrong place for them: it bloats the persisted state and there's no sane size limit on it. So it isn't extra data we'd be choosing to store — it's data we're already storing, just somewhere much worse. That's also why it can't be opt-in: until it syncs, that recording is the *only* copy of the user's data.

That work is for later PRs, not this one. But two things are much cheaper to do **before** this merges, while the schema isn't in the wild yet:

- rename the DB `cboard-image-cache` → `cboard-media`, and the store `images` → `cached` (the `byLastUsed` index and the `meta.totalBytes` accounting stay exactly as they are)
- create the empty `local` store in the same v1 `upgrade`, even though nothing writes to it yet

Keeping them as two separate stores is what actually answers your space worry, I think. The two have opposite lifetimes: a cached symbol can always be re-fetched, so LRU eviction is fine and it's free to use whatever budget is left; a sound recorded offline must never be evictable. Separate stores means local media can't be evicted by the cache, and the cache can't grow at the expense of unsynced user data. Same DB, different rules.

Your ranking still holds inside the cache itself, for what it's worth — a missing symbol breaks the board, a missing sound degrades to TTS.

eturino:
@RodriSanchez1 Done 👍 

> That work is for later PRs, not this one. But two things are much cheaper to do **before** this merges, while the schema isn't in the wild yet:
> 
> * rename the DB `cboard-image-cache` → `cboard-media`, and the store `images` → `cached` (the `byLastUsed` index and the `meta.totalBytes` accounting stay exactly as they are)
> * create the empty `local` store in the same v1 `upgrade`, even though nothing writes to it yet
>

eturino:
@RodriSanchez1 pinging again (sorry) 😅

RodriSanchez1:
Hi @eturino! Sorry for the delay — I got back to this today. I have a few tweaks ready: do you mind if I commit them here directly, or would you prefer a separate PR?

Two things I ran into:

**1. CORS on Global Symbols images.** Their API sends `Access-Control-Allow-Origin: *` in most of the images , but there are someones that s don't:

GET /symbolsets/arasaac/symbols/16359.png → (no ACAO header)

https://globalsymbols.com/symbolsets/arasaac/symbols/16359.png?locale=en


So on the web the `fetch()` is blocked and the image never reaches IndexedDB — the `` still renders, but the cache stays empty. Android and iOS are fine, since the webviews don't enforce CORS. Worth opening an issue upstream: it's a one-line change on their side, and it would fix the web build for free.

Global Symbols recently changed their api. This is issue that you are experience started lately?

**2. Images blink during board navigation.** Noticeable on big boards with cached images: the browser cache paints the URL first, then the IndexedDB lookup swaps in a blob and the whole grid flickers. 

I'm working on this!

eturino:
commit ahead of course, I'll pull your changes and check out in local. More than happy to share the progress

RodriSanchez1:
Nice! I'm currently in touch with the Global Symbols developer team about the image caching issue, and we're waiting on a fix from their side.

**The problem:** some `picto.image_url` (that is what we store in `tile.image`) values return a `302` into `/symbolsets/`, and the redirected response is missing the headers we need:


```
GET /uploads/production/image/imagefile/19232/17_19233_....png
→ 302 Found
Access-Control-Allow-Origin: *
Location: /symbolsets/arasaac/symbols/19233.png?locale=en-GB
→ 200 OK
Cache-Control: no-cache
-- no Access-Control-Allow-Origin
-- no ETag, no Last-Modified
```

Browsers apply the CORS check to **every** response in a redirect chain, so the header on the `302` doesn't carry over — the read is blocked at the final hop. Combined with `Cache-Control: no-cache` and no validator, these images can't be stored in our on-device cache and get re-downloaded on every load.

**Impact:** affected symbols render fine while online, but the tiles go blank as soon as the user is offline — on web and on the packaged Android build.

When they fix it, it supposed that the app will return working as expected. Maybe you can confirm that the losted images in your son board does this redirection. 

Anyways I think this idb solution gives us a good fallback image render instead on trusting on the  http cache

RodriSanchez1 (approved):

PR #2246, opened 2026-06-29
fix(board): sanitize base64 tile sounds before board create/update

### LINKED ISSUE #2245: Base64 recorded tile sounds bypass board-media sanitization on sync
### Summary

Recorded tile sounds are stored as base64 `data:` URLs (`VoiceRecorder` uses `FileReader.readAsDataURL`). PR #2242 added sanitization that uploads local-only tile **images** to `/media` before a board is created/updated on sync, but it only inspects `tile.image` — `tile.sound` is left untouched.

As a result a base64 `tile.sound` can still reach the server on sync, which is the same class of failure #2242 fixed for images: the server expects a URL (the existing `Board.container.js#uploadTileSound` already uploads base64 sound to `/media` and swaps in the URL), so a base64 `sound` causes the board create/update to be rejected and the board's `syncMeta` status stays `PENDING`, retried every cycle.

### How it reaches the server

`Board.container.js#uploadTileSound` only runs in the interactive edit path (`handleApiUpdates`), gated behind a logged-in user. A base64 sound survives when it bypasses that path, e.g.:

- a sound recorded while offline / not logged in, later synced, or
- `uploadTileSound` failing (its `catch` swallows the error).

The sync chokepoint (`createApiBoard` / `updateApiBoard`, where all sync sub-paths funnel through) sanitizes the image but not the sound.

### Expected

The board-media sanitizer at the create/update chokepoint should also upload base64 `tile.sound` to `/media` and replace it with the resulting URL, with the same partial-success / defer-on-failure semantics #2242 established for images.

### Related

- #2242 (image sanitization this extends)

RodriSanchez1:
## Summary

Follow-up to #2242. That PR added a board-media sanitizer that uploads local-only tile **images** to `/media` before a board is created/updated on sync. It only inspected `tile.image`, so base64 **sounds** were left unhandled.

Recorded tile sounds are stored as base64 `data:` URLs (`VoiceRecorder` uses `FileReader.readAsDataURL`). A base64 `tile.sound` can reach the server on sync — the same failure class #2242 fixed for images — leaving the board `PENDING` and retried every cycle. The existing `Board.container.js#uploadTileSound` only converts sound on the interactive, logged-in edit path; anything that bypasses it (sound recorded offline / not logged in, or `uploadTileSound` failing silently) slips through.

Fixes #2245.

## Changes

- **`refactor(board)`** — rename `uploadBoardLocalImages` → `uploadBoardLocalMedia` and `sanitizeBoardImages` → `sanitizeBoardMedia` (pure rename, no behavior change).
- **`fix(board)`** — extend the sanitizer to also upload base64 `tile.sound` to `/media` and swap in the URL, with the same partial-success / defer-on-failure semantics used for images. Per-tile image and sound uploads run in parallel via dedicated `uploadTileImageMedia` / `uploadTileSoundMedia` helpers; the orchestrator keeps the 5-tile concurrency batching, `hadFailure` aggregation, and board reconstruction.

Sound is only ever a base64 `data:` URL (recordings are never `file://`), so it uses the data-URL path only.

## Behavior on failure (unchanged from #2242)

- A single failed upload leaves that field un-replaced and sets `hadFailure`; other tiles still process.
- Partial successes are committed locally via `updateBoard` before deferring.
- If anything failed, the board's server create/update is skipped this cycle and stays `PENDING`, retried next cycle.

## Testing

- `uploadBoardLocalMedia`: added cases for base64 sound replacement, a tile carrying both image + sound, and sound-upload failure flagging `hadFailure`.
- `sanitizeBoardMedia`: added a sanitized-sound dispatch case.
- New media-sanitizer tests pass locally; the pre-existing `axios`-mock cases in `api.test.js` are unchanged (the cordova-util mock is left identical to master).

## Notes

- The redundant `Board.container.js#uploadTileSound` fast-path is left in place (immediate local-state update); the chokepoint sanitizer now acts as the safety net. Possible future consolidation.
- Assumes the server rejects base64 `sound` like it rejects base64 `image` — the existence of `uploadTileSound` is the evidence.

### DIFF
--- a/src/api/api.js
+++ b/src/api/api.js
@@ -464,47 +464,81 @@ class API {
     return url;
   }
 
+  async uploadTileImageMedia(tile) {
+    if (isDataURL(tile.image)) {
+      const url = await this.uploadFromDataURL(tile.image, tile.id, true);
+      return { attempted: true, url };
+    }
+
+    if (isLocalFileURL(tile.image) && isAndroid()) {
+      const file = await new Promise(resolve => {
+        window.resolveLocalFileSystemURL(
+          tile.image,
+          fileEntry => {
+            fileEntry.file(file => resolve(file), () => resolve(null));
+          },
+          () => resolve(null)
+        );
+      });
+      if (file) {
+        const segments = tile.image.split('/');
+        const name = segments[segments.length - 1] || tile.id;
+        const url = await this.uploadFile(file, name);
+        return { attempted: true, url };
+      }
+      return { attempted: true, url: null };
+    }
+
+    return { attempted: false, url: null };
+  }
+
+  async uploadTileSoundMedia(tile) {
+    if (isDataURL(tile.sound)) {
+      const url = await this.uploadFromDataURL(tile.sound, `${tile.id}.mp3`);
+      return { attempted: true, url };
+    }
+
+    return { attempted: false, url: null };
+  }
+
   async uploadBoardLocalMedia(board) {
     const tiles = board?.tiles || [];
     const targets = tiles.filter(
-      tile => isDataURL(tile?.image) || isLocalFileURL(tile?.image)
+      tile =>
+        isDataURL(tile?.image) ||
+        isLocalFileURL(tile?.image) ||
+        isDataURL(tile?.sound)
     );
 
     if (!targets.length) {
       return { board, hadFailure: false };
     }
 
-    const urlByTileId = {};
+    const imageUrlByTileId = {};
+    const soundUrlByTileId = {};
     let hadFailure = false;
 
     const uploadTarget = async tile => {
       try {
-        let url = null;
-        if (isDataURL(tile.image)) {
-          url = await this.uploadFromDataURL(tile.image, tile.id, true);
-        } else if (isLocalFileURL(tile.image) && isAndroid()) {
-          const file = await new Promise(resolve => {
-            window.resolveLocalFileSystemURL(
-              tile.image,
-              fileEntry => {
-                fileEntry.file(file => resolve(file), () => resolve(null));
-              },
-              () => resolve(null)
-            );
-          });
-          if (file) {
-            const segments = tile.image.split('/');
-            const name = segments[segments.length - 1] || tile.id;
-            url = await this.uploadFile(file, name);
+        const [image, sound] = await Promise.all([
+          this.uploadTileImageMedia(tile),
+          this.uploadTileSoundMedia(tile)
+        ]);
+
+        if (image.attempted) {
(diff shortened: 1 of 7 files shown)


src/api/api.js
```diff
@@ -464,47 +464,81 @@ class API {
...
+      }
+      return { attempted: true, url: null };
+    }
+
+    return { attempted: false, url: null };
+  }
+
+  async uploadTileSoundMedia(tile) {
+    if (isDataURL(tile.sound)) {
+      const url = await this.uploadFromDataURL(tile.sound, `${tile.id}.mp3`);
```
magush27, line 497: Here VoiceRecorder uses MediaRecorder, whose output format depends on the browser (audio/webm on Chrome, audio/ogg on Firefox). The file extension .mp3 will be wrong for WebM or OGG recordings. This could cause playback issues depending on how the media server serves files or how the browser determines MIME type. Worth checking what format VoiceRecorder actually produces and if the server infers MIME from the data URL's prefix or from the filename.
There's also a deeper inconsistency in `uploadFromDataURL` that passes `checkExtension` to `dataURLtoFile` (which builds the correct `file.name` internally), but then calls `uploadFile(file, filename)`  this is the original `filename`, not `file.name`. So `checkExtension` is actually a no-op for anything uploaded through this path. Images work because the server doesn't appear to care about filename extension (it uses MIME type).

A suggestion on how to fix is to extract the extension from the data URL directly in `uploadTileSoundMedia`, keeping it self-contained and not touching the image path:

```
  async uploadTileSoundMedia(tile) {
    if (isDataURL(tile.sound)) {
      const mime = tile.sound.match(/^data:(audio\/[^;,]+)/)?.[1] ?? 'audio/mpeg';
      const ext = mime.split('/')[1] || 'mp3';
      const url = await this.uploadFromDataURL(tile.sound, `${tile.id}.${ext}`);
      return { attempted: true, url };
    }
    return { attempted: false, url: null };
  }
```
This produces `tile.id.webm` on Chrome, `tile.id.ogg` on Firefox, `tile.id.mp4` on Safari, the correct extension in every case.

And also this other fix:
```
async uploadFromDataURL(dataURL, filename, checkExtension = false) {
    const file = dataURLtoFile(dataURL, filename, checkExtension);
    return await this.uploadFile(file, file.name); // use the computed name, not the raw filename
  }
```

magush27 (commented):
I'd been testing it and it works great!
The only thing I found is that the `shouldUpdate` path that now calls `uploadBoardLocalMedia` before `updateBoard` has no new tests, a test covering it would be a good to add.

PR #2254, opened 2026-07-01
fix(media): reliable Android tile media upload + stop PENDING-forever boards

### LINKED ISSUE #2253: Android: tile images/sounds fail to upload and boards get stuck in PENDING sync
## Problem

On Android (Cordova WebView), tile media created/edited offline fails to upload during sync, and the affected board never leaves the `PENDING` sync state — it retries forever.

Two distinct root causes:

### 1. Media uploaded as a reconstructed `File` → `status 0` / 0 bytes
When a tile image/sound was stored as base64 (`data:`) or a local `file://`, the sanitizer built a `File` from a `Uint8Array` (or passed a Cordova `File` from `fileEntry.file()`) straight into `FormData`. The Android System WebView does not serialize those into the multipart body, so the POST to `/media` reaches the server empty (surfaced as `status 0`). The online path worked because it uploads a real `Blob` (from `canvas.toBlob`).

Note: `fetch()` of a `data:` URL is **not** an option — it's blocked by the app CSP (`connect-src` has no `data:`).

### 2. Boards stuck in PENDING forever
When media can **never** be retrieved on the current device — e.g. a `file://` that belongs to another device and doesn't resolve, or a base64 that can't be decoded — the sanitizer flagged a failure, which aborts the board create/update and keeps it `PENDING`. Every subsequent sync retries the same impossible upload.

## Expected

- Local media (base64 and `file://`) uploads reliably on Android.
- Media that genuinely cannot be retrieved is dropped (replaced with an empty string) so the board can sync.
- Real transient upload/network failures still retry (board stays PENDING until it succeeds).

## Fix

- Decode base64 into a real `Blob` (`dataURLtoBlob`, CSP-safe) and read `file://` bytes via `FileReader` before upload.
- Distinguish **unrecoverable** (can't obtain the bytes → clear to `''`, don't flag failure) from **transient** (bytes obtained but POST failed → keep retrying).
- Stop generating new `file://` tile images; centralize CDN URL rewriting in `uploadFile`.

Fixed in PR from branch `fix/android-media-upload`.

RodriSanchez1:
Closes #2253

## What

Fixes two Android media-sync problems:

1. **Uploads fail with `status 0` / 0 bytes.** Tile media stored as base64 (`data:`) or `file://` was uploaded as a `File` reconstructed from a `Uint8Array` (or a raw Cordova `File`), which the Android WebView does not serialize into the multipart body. Now everything is uploaded as a real `Blob`:
   - base64 → `dataURLtoBlob` (decode with `atob`, CSP-safe — no `fetch` of `data:` URLs, which the CSP blocks).
   - `file://` → read bytes via `FileReader.readAsArrayBuffer` → `Blob`.
2. **Boards stuck in `PENDING` forever.** Media that can never be retrieved on this device (a `file://` from another device that doesn't resolve/read, or an undecodable base64) is now replaced with `''` instead of flagging a failure, so the board syncs. Genuine transient upload/network failures still keep the board `PENDING` and retry.

## How

- `tryUploadDataURL` splits decode (unrecoverable) from upload (transient) and returns `{ url, unrecoverable }`.
- `uploadTileImageMedia` / `uploadTileSoundMedia` thread the `unrecoverable` flag; `file://` resolve/read failures are unrecoverable, POST failures are transient.
- `uploadBoardLocalMedia` clears unrecoverable media to `''` (no `hadFailure`) and only flags `hadFailure` for transient failures.
- `updateTileImgURL` (TileEditor) simplified: no longer generates new `file://` images; always falls back to base64 when not logged in.
- CDN URL rewriting (`convertMediaUrlToCDN`) centralized in `uploadFile`, so images **and** sounds pass through it.

## Notes

- The `file://` read branch is kept as a migration path for tiles saved by older app versions.
- Tests updated in `api.test.js` (mock `tryUploadDataURL`; new cases: undecodable base64 → `''`, unresolvable `file://` → `''`, both without `hadFailure`).

## Follow-up (out of scope)

- Persistent server rejection of valid bytes still loops as PENDING — a retry cap could be added later.
- `Board.container.uploadTileSound` duplicates the base64→Blob sound-upload path; consolidate on `dataURLtoBlob` (tech debt).

### DIFF
--- a/src/__test__/helpers.test.js
+++ b/src/__test__/helpers.test.js
@@ -1,13 +1,9 @@
 import * as helpers from '../helpers';
 describe('helpers', () => {
-  it('should create a file from data', () => {
+  it('should create a blob from a data URL', () => {
     const dataUrl = 'data:text/plain;charset=utf-8;base64,dGVzdGluZw==';
-    const file = helpers.dataURLtoFile(dataUrl, 'myfile');
-    expect(file.name).toBe('myfile');
-  });
-  it('should create a file with ext from data', () => {
-    const dataUrl = 'data:text/plain;charset=utf-8;base64,dGVzdGluZw==';
-    const file = helpers.dataURLtoFile(dataUrl, 'myfile.txt', true);
-    expect(file.name).toBe('myfile.txt.plain');
+    const blob = helpers.dataURLtoBlob(dataUrl);
+    expect(blob.type).toBe('text/plain');
+    expect(blob.size).toBe(7);
   });
 });
--- a/src/api/api.js
+++ b/src/api/api.js
@@ -20,6 +20,12 @@ import { cvaFileToBlob, isAndroid } from '../cordova-util';
 
 const BASE_URL = API_URL;
 const LOCAL_COMMUNICATOR_ID = 'cboard_default';
+
+const FILE_NOT_FOUND_ERR = 1;
+const FILE_ENCODING_ERR = 5;
+const isUnrecoverableFileError = error =>
+  !!error &&
+  (error.code === FILE_NOT_FOUND_ERR || error.code === FILE_ENCODING_ERR);
 export let improvePhraseAbortController;
 
 const getUserData = () => {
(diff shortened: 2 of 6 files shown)


tomivm:
One thing worth a second look before merge: in `uploadTileImageMedia` (`src/api/api.js`), a `resolveLocalFileSystemURL` resolve/read failure now returns `unrecoverable: true`, which clears `tile.image` to `''` and syncs immediately.

Before this PR, that same failure set `hadFailure = true`, so the board stayed `PENDING` and retried, keeping the local image intact. The resolve callback's error path can fire for reasons that aren't "file is really gone" (permission race right after an app update, storage momentarily unmounted, a cold-start provider race), so this turns "retry and keep" into "delete permanently" for a wider set of cases than just genuinely missing files.

Is that trade-off intentional? If not, might be worth keeping resolve/read errors in the `hadFailure`/retry path, and reserving the `unrecoverable`/clear-to-empty behavior for a confirmed not-found case.

PR #2271, opened 2026-07-14
Fix deep linking on mobile (Cordova/Ionic)

magush27:
This PR fixes deep link navigation not working correctly in the Cordova/Android build.

**Changes:**
- `src/index.js`: Replaced `HashRouter` with a `Router` using the shared history singleton for Cordova builds. This ensures deep link navigation and programmatic `history.push()` calls affect the same router instance.
- `src/cordova-util.js`: Fixed the Ionic deep link route matching `match.$route` returns the route object (with a target property), not a string, so the condition was never matching. Updated to use `match.$route.target` instead.

**How to test it:**
Aside from the changes you need to make in the code to build the test APK explained in the README.md, you need to change the file `constants.js` in line 9 at `RAW_API_URL` to:

```
const RAW_API_URL = isCordova()
  ? 'https://api.app.qa.cboard.io/'
  : DEV_API_URL || `${window.location.protocol}//api.${host}`;
```

Also, in the `ccboard` repo, you need to change `package.json`  ionic-plugins-deeplink constant `DEEPLINK_HOST` to:
```
"DEEPLINK_HOST": "app.qa.cboard.io",
```

Also in `platforms/android/src/main/AndroidManifest.xml` you need to replace the lines that has `app.cboard.io` url to `app.qa.cboard.io`. 

Once done, you can't build the APK using the classic `cordova build android` because it rewrites the manifest using the production links. You need to go to `platforms/android` and there run `./gradlew assembleRelease`. The APK will be at:

```
  platforms/android/app/build/outputs/apk/release/app-release.apk
```

Once you install the APK in your device, you should activate the app links for `app.qa.cboard.io` because debug apps doesn't open them directly as PlayStore apps do. Using ADB, you can run:
```
  adb shell pm set-app-link --package com.unicef.cboard always
```
After either option, any link to https://app.qa.cboard.io/... tapped in a browser or message app will open the APK directly instead of the browser.

This let you test the app with QA link.


src/index.js
```diff
@@ -30,8 +36,10 @@ const dndOptions = {
   enableKeyboardEvents: true
 };
 
-// When running in Cordova, must use the HashRouter
-const PlatformRouter = isCordova() ? HashRouter : BrowserRouter;
+// When running in Cordova, use Router with shared history so deep linking works
+const PlatformRouter = isCordova()
+  ? ({ children }) => <Router history={history}>{children}</Router>
+  : BrowserRouter;
```
RodriSanchez1: **Suggestion: drop `PlatformRouter` entirely.**

The fix is right, but it only reaches half the app. `src/history.js` already picks the right history for the platform:

```js
const history = isCordova() ? createHashHistory() : createBrowserHistory();
```

So by the time `index.js` imports `history`, the Cordova-vs-web decision has already been made. The `isCordova()` ternary here makes it a second time, and the two branches now disagree about *which* history the Router listens to: Cordova gets the shared singleton, web gets a private one that `BrowserRouter` builds for itself.

That private instance is the same bug this PR is fixing, just on the other platform. Anything that calls `history.push()` from outside a component — `initDeepLinking`, the 401 handler in `api.js:84`, the `history.replace()` calls in `Board.actions.js` and `Communicator.actions.js` — writes to the singleton, and on web the Router never hears about it. (It has been mostly invisible so far because those board actions pair the `replace` with a Redux dispatch that re-renders the screen anyway, and because a hash change fires a global `hashchange` event that a `HashRouter` *does* pick up. Neither safety net exists for `BrowserRouter`.)

Since both platforms want "use the shared history", the conditional can just go away:

```js
import { Router, Route } from 'react-router-dom';
import history from './history';

// history.js already selects hash history under Cordova, browser history on web
<Router history={history}>
  ...
</Router>
```

One history instance, one Router listening to it, and the platform decision stays in the one file that owns it. It also removes the inline arrow component — safe as written since it is evaluated once at module scope, but if that expression ever drifts into a render body it becomes a fresh component type on every pass and remounts the whole tree.

magush27:
Closed it because the merge branch was wrong, the correct PR is this one https://github.com/magush27/cboard/pull/2

PR #2321, opened 2026-08-14
fix: use hash history in packaged apps to avoid 404 page after splash

martinbedouret:
## Problem

After the API URL fix (#2320), the Electron, iOS and Android apps show the **404 (NotFound) page right after the splash screen**.

## Root cause

Same `isCordova()` timing bug as the API URL, this time in `src/history.js`:

```js
const history = isCordova() ? createHashHistory() : createBrowserHistory();
```

`isCordova()` is `!!window.cordova`, but `history.js` runs at **module-evaluation time**, before Cordova defines `window.cordova`. So packaged builds get `createBrowserHistory()` (HTML5 path-based history). Packaged apps are served over non-web schemes:

- Electron → `file://`
- Android → `file://` (`AndroidInsecureFileModeEnabled=true`)
- iOS → `app://localhost`

Under those schemes, HTML5 history paths (e.g. `/board/…`) don't resolve, so react-router matches nothing and renders `<Route component={NotFound} />` — the 404 page.

## Fix

Select the history implementation from the **actual location scheme** instead of the not-yet-ready `window.cordova`:

- `http:` / `https:` → `createBrowserHistory()` (web, including local dev — unchanged)
- everything else (`file:`, `app:`) → `createHashHistory()` (packaged apps)

This removes the dependency on Cordova initialization timing and mirrors the approach used for the API URL fix.

## Testing

- Rebuilt and repackaged the Electron app: it now loads past the splash into the app instead of the 404 page; navigation works via hash routes.
- Web behavior unchanged: `http`/`https` still use browser history (production and local dev).

## Related

Follow-up to #2320 (same underlying `isCordova()`-at-module-load issue).

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -38,11 +38,11 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       # Initializes the CodeQL tools for scanning.
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@v2
+        uses: github/codeql-action/init@v3
         with:
           languages: ${{ matrix.language }}
           # If you wish to specify custom queries, you can do so here or in a config file.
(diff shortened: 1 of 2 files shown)


martinbedouret:
Superseded by #2322, which includes this history fix refactored onto the new shared `isPackagedApp()` helper (along with the API URL and broken-image fixes). Closing in favor of #2322.

PR #2347, opened 2026-09-17
fix: do not paint a broken symbol image while offline

RodriSanchez1:
## Problem

After #2244 (remote symbol images cached in IndexedDB for offline use), every symbol on a board showed the browser's broken-image glyph while offline, until the IndexedDB lookup answered and swapped in the stored copy.

Root cause: `SymbolImage` painted the tile's remote url first and only read storage from `onError`. The one effect that read storage proactively was gated `if (!image && keyPath)`, which never fires for real tiles — `SymbolSearch.component.js` saves both `image` and `keyPath`. So nothing touched IndexedDB until the network request had already failed, and the glyph was on screen for the whole lookup. It repeated on every mount (`key={image || keyPath}`, plus fresh mounts per board navigation, per FixedGrid page and per Output push).

## Fix

A *pending* `` paints nothing; only a *failed* one paints the glyph. So offline the remote url stays the `src` and the image is rendered with `visibility: hidden` until `onLoad` proves it loaded; `onError` now clears `src` before falling back to storage, and the storage-read effect is gated on `!src` instead of the condition that never fired.

## Why url-first online is kept

An `` http-cache hit is decoded off the main thread with no JS, no Blob, no `createObjectURL` and no React re-render. The IndexedDB path costs all of those *plus* the same decode. This change touches only the offline branch — online there is still no lookup and no extra render — and it keeps the renderer memory cache on the fast path, so a board rendered earlier in the same session still paints instantly when revisited offline.

An earlier attempt that withheld the url offline and resolved IndexedDB before painting was rejected: it skipped the renderer memory cache and could not cover a flaky network, where `navigator.onLine` stays `true`.

## Tests

`Symbol.test.js` gains offline/visibility cases: hidden until `onLoad` offline, visible immediately online, fallback to the stored image on error, and no lookup while online.

## Not in this PR

- A module-level url memo so repeat mounts skip IndexedDB entirely — deferred by request to its own PR, tracked in a follow-up issue.
- A tile that ends blank offline with nothing stored does not recover when the network returns (`src` no longer changes, so `onError` cannot re-fire). Pre-existing behaviour — it stayed broken rather than blank — and left for the memo PR to decide.

https://claude.ai/code/session_01PSVTZoXRHXRkSc9RZMCaNP


RodriSanchez1:
Follow-up tracked in #2348 (session url memo so repeat mounts skip IndexedDB), deliberately out of scope here.

